PR #192, opened 2018-10-12
Add Razor serialization converters.

### LINKED ISSUE #184: Razor errors for non-ASP.NET Core projects
#### Is this a Bug or Feature request?:
Bug

#### Steps to reproduce
Open a non-ASP.NET Core project that contains Razor files.

#### Description of the problem:

Expected behavior:
No experience

Actual behavior:
image

#### Additional information

This issue is caused because we don't know what version of Razor to parse old asp.net core projects. Therefore, we generate invalid code that doesn't work with the users project.

NTaylorMullen:
- These converters are nearly direct copies from the Razor repo and are necessary to serialize project level information.

Need sign off in order to merge these guys given the new PR requirements. This PR is just prep work for bigger changes that I'm going to be pushing through to fix #184.


NTaylorMullen:
> So, my only feedback is: Is there any value in trying to share these sources with whatever else defines them? Is it worth moving these to one of the shared-sources common repos?

I had the same idea as well. This is something we might take on once I'm back from vaca but for now I'm hacking together solutions to try and get the legacy Razor projects fix in by Friday.

PR #195, opened 2018-10-15
Add OmniSharp plugin infrastructure to understand Razor projects.

NTaylorMullen:
- This plugin infrastructure works off of MEF and consumes a `RazorConfigurationProvider`. That provider gets the project instance and has a chance to tell the system it does or does not understand a project. First one wins. We may eventually need to add ordering to the providers but we can do that when the time comes.
- To test this added support for ASP.NET Core Razor versions < 2.1 and >= 2.1.
- Added tests to verify all of the infrastructure pieces.
- Added TargetFramework to the project.razor.json file for future changes.

#184


src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/RazorConfigurationProvider.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Razor.Language;
+
+namespace Microsoft.AspNetCore.Razor.OmniSharpPlugin
+{
+    public abstract class RazorConfigurationProvider
```
NTaylorMullen: Blazor will  eventually provide one of these guys.

test/testapps/BasicRazorApp1_0/Views/Index.cshtml
```diff
@@ -0,0 +1,7 @@
+﻿@page
```
NTaylorMullen: Nuke in the 1.0 app

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/DefaultConfigurationProvider.cs
```diff
@@ -0,0 +1,222 @@
...
+using System.Composition;
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.Build.Execution;
+
+namespace Microsoft.AspNetCore.Razor.OmniSharpPlugin
+{
+    [Shared]
+    [Export(typeof(RazorConfigurationProvider))]
+    internal class DefaultConfigurationProvider : CoreProjectConfigurationProvider
```
SteveSandersonMS: From the logic, it appears that `DefaultConfigurationProvider` is for >2.1, and `FallbackConfigurationProvider` is for <=2.1. The Default/Fallback naming doesn't imply versioning strongly to me - would names like Current vs Legacy be clearer? Or even indicate which side of the 2.2 boundary they are on explicitly in the name?

NTaylorMullen: Oops, should update the comment. Default is for >= 2.1 and Fallback is for < 2.1. As for naming, I brought them forward  from the Razor repo. Our project system impls which do similar things to these use those names; I think you're right though, we can do better 👍

Issue #197, opened 2018-10-15
Add 1.0 Razor integration test


NTaylorMullen:
Once https://github.com/aspnet/Razor.VSCode/issues/168 has been completed we can add 1.0 Razor integration tests that verify we parse pre-2.1 projects correctly. This is also dependent on #184 being completed.

NTaylorMullen:
a3647cd9fd444791246d47ff7745a2dbca908bb7

PR #198, opened 2018-10-15
Add Razor project management to client.

NTaylorMullen:
- Changed flow of project resolution to be the following: 1. Discover/listen for all projects (without their configuration) 2. Discover/listen for project data files. With this flow we can always support projects that haven't been restored, don't have valid data or are slower to initialize. This required some changes to how we add projects.
- Added client pieces that listen and look for project.razor.json files and dispatch notifications.
- Updated ProjectTracker to listen to the project manager and notify server of project changes.
- Grouped project.razor.json files by tfm on client to differentiate them. We don't currently care about TFMs but for correctness we have them for understandability.
- Added server side tests and filed an issue to create integration tests to validate the new multi-targeting behavior #197
- Changed the AddProject signature to no longer take a configuration.

#184


src/Microsoft.AspNetCore.Razor.VSCode/src/RPC/UpdateProjectRequest.ts
```diff
@@ -0,0 +1,20 @@
+/* --------------------------------------------------------------------------------------------
+* Copyright (c) Microsoft Corporation. All rights reserved.
+* Licensed under the MIT License. See License.txt in the project root for license information.
+* ------------------------------------------------------------------------------------------ */
+
+import { IRazorProjectConfiguration } from '../IRazorProjectConfiguration';
+
+export class UpdateProjectRequest {
```
SteveSandersonMS: I don't really understand why this class is defined. It seems to be a repeat of the `IRazorProjectConfiguration` type. I can see that the names of the `projectFilePath`/`projectPath` property differ, but is there any other reason for this to exist instead of using `IRazorProjectConfiguration` directly everywhere?

SteveSandersonMS: Update: No forget that comment, I see that `UpdateProjectRequest` is a small subset of what is on `IRazorProjectConfiguration`.

Arguably it would be clearer if `UpdateProjectRequest` was defined as an interface too, since it doesn't contain any logic and is just a DTO. Not critical though!

src/Microsoft.AspNetCore.Razor.VSCode/src/RazorProjectManager.ts
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        if (!this.razorProjects[configuration.projectPath]) {
+            this.logger.logVerbose(
+                `Invalid project config. Could not find a corresponding project for ${configuration.projectPath}`);
+            return;
+        }
+
+        const projectContainer = this._getProject(configuration.projectUri);
+        const newConfiguration = { ...projectContainer.configuration };
```
SteveSandersonMS: Could you clarify why it's necessary to clone the earlier `configuration` dict before editing it? It looks like we're overwriting the old `this.razorProjects` entry entirely anyway, so I'm not sure why we can't just mutate it in place. Are the `configuration` objects somehow shared by multiple projects sometimes?

src/Microsoft.AspNetCore.Razor.VSCode/src/RazorProject.ts
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    public getCurrentConfiguration() {
+        let latestConfiguration: IRazorProjectConfiguration | undefined;
+        const allConfigurations = Object.values(this.configuration);
+        for (const projectData of allConfigurations) {
+            if (!latestConfiguration || latestConfiguration.lastUpdated < projectData.lastUpdated) {
+                latestConfiguration = projectData;
+            }
+        }
```
SteveSandersonMS: I haven't spotted what value we get from tracking multiple configurations (keyed by TFM). It looks like we only ever use the most recently written one. Is there reason to think this will be needed later? If not then maybe there's a possible quick simplification.

src/Microsoft.AspNetCore.Razor.VSCode/src/RazorProjectManager.ts
NTaylorMullen: We could mutate it in place; however, i'm trying to correct one of the mistakes I made with the `RazorDocumentManager`. Having immutable projects enforces consumers to rely on the change event from this class rather than holding onto the projects. That way we have less coupling and have more flexibility on what we can and cannot change in regards to the Razor projects.

This shouldn't happen frequently so it's safer to make things immutable.

src/Microsoft.AspNetCore.Razor.VSCode/src/RazorProject.ts
NTaylorMullen: Ya good point, I was going to do more with the multiple configs but then went back on that idea. I'll change it to only track 1.

src/Microsoft.AspNetCore.Razor.VSCode/src/IRazorProjectConfiguration.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import * as vscode from 'vscode';
+
+export interface IRazorProjectConfiguration {
+    readonly path: string;
+    readonly uri: vscode.Uri;
+    readonly projectPath: string;
+    readonly projectUri: vscode.Uri;
+    readonly tagHelpers: any[];
+    readonly configuration: any;
```
SteveSandersonMS: Do we only use this `configuration` data to pass through to the language server without inspecting it on the JS side? If so, totally reasonable to type it as `any`. If we do need to process it within TypeScript though it would be nice to have a type definition for it.

NTaylorMullen: Mostly pass through. The only other thing that will be inspecting it will be the telemetry bits to report on project types.

NTaylorMullen: I'll address that in the telemetry PR.

PR #199, opened 2018-10-16
Don't generate C# for SystemWeb projects.

NTaylorMullen:
- This involved creating an Unsupported Razor configuration that adds a disclaimer to the generated Razor C# but refuses to do any other calculations.
- Added some extensions to allow us to understand when we're operating in an "unsupported" scenario. This means the server's API endpoints have some knowledge of being unsupported but the client code doesn't need to know anything. For instance, our language query endpoint forces Html always when working with unsupported documents; this is consistent with how VSCode used to operate.
- Added a SystemWeb plugin component that looks for `System.Web.Razor.dll` and uses that to determine if a project targets the old ASP.NET world. In the case that it does the SystemWeb plugin piece sets the project configuration to Unsupported.
- Added tests for all of the various components that were added.

#184

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/SystemWebConfigurationProvider.cs
+++ b/src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/SystemWebConfigurationProvider.cs
@@ -1,0 +1,56 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Composition;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.AspNetCore.Razor.LanguageServer;
+
+namespace Microsoft.AspNetCore.Razor.OmniSharpPlugin
+{
+    [Shared]
+    [Export(typeof(RazorConfigurationProvider))]
+    internal class SystemWebConfigurationProvider : CoreProjectConfigurationProvider
+    {
+        // Internal for testing
+        internal const string ReferencePathWithRefAssembliesItemType = "ReferencePathWithRefAssemblies";
+        internal const string SystemWebRazorAssemblyFileName = "System.Web.Razor.dll";
+
+        private const string LegacyRazorAssemblyName = "System.Web.Razor";
+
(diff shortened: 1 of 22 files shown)


SteveSandersonMS:
Cool. I’ll do a pass through reviewing the open PRs in a couple of hours.

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/SystemWebConfigurationProvider.cs
```diff
@@ -0,0 +1,51 @@
...
+            {
+                var assemblyPath = compilationReference.EvaluatedInclude;
+                if (compilationReference.EvaluatedInclude.EndsWith(SystemWebRazorAssemblyFileName, StringComparison.OrdinalIgnoreCase))
+                {
+                    legacyRazorReferenceFullPath = compilationReference.EvaluatedInclude;
+                    break;
+                }
+            }
+
+            if (legacyRazorReferenceFullPath == null)
```
NTaylorMullen: legacy => `systemWeb`

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/SystemWebConfigurationProvider.cs
```diff
@@ -0,0 +1,51 @@
...
+                configuration = null;
+                return false;
+            }
+
+            var compilationReferences = context.ProjectInstance.GetItems(ReferencePathWithRefAssembliesItemType);
+            string legacyRazorReferenceFullPath = null;
+            foreach (var compilationReference in compilationReferences)
+            {
+                var assemblyPath = compilationReference.EvaluatedInclude;
+                if (compilationReference.EvaluatedInclude.EndsWith(SystemWebRazorAssemblyFileName, StringComparison.OrdinalIgnoreCase))
```
SteveSandersonMS: I don't know if `compilationReference.EvaluatedInclude` is a path string here, but if it is, maybe we need the check to only match if there's a directory separator before `SystemWebRazorAssemblyFileName` (to avoid matching `TaylorSystem.Web.Razor.dll`).

NTaylorMullen: Good catch, ya i'll definitely do that.

Issue #194, opened 2018-10-15
Add Razor project configuration telemetry


NTaylorMullen:
#### Is this a Bug or Feature request?:
Feature

#### Description of the problem:
We need to know the common project types that are being used with tooling in order to prioritize features.

NTaylorMullen:
Telemetry bits have been submitted. We'll know if they're approved Tuesday.

PR #200, opened 2018-10-17
Report project info in telemetry.

NTaylorMullen:
- Add a `ProjectTelemetryListener` which listens for `RazorProjectManager` changes and reports telemetry events.
- Reported telemetry is only reported for significant project changes that result in Razor's project configurations bieng significantly different.
- Update extension shim to log additional telemetry event data.

#194

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.VSCode/src/TelemetryReporter.ts
+++ b/src/Microsoft.AspNetCore.Razor.VSCode/src/TelemetryReporter.ts
@@ -5,9 +5,11 @@
 
 import { Trace } from 'vscode-jsonrpc';
 import { HostEventStream, TelemetryEvent } from './HostEventStream';
+import { IRazorProject } from './IRazorProject';
 
 export class TelemetryReporter {
     private readonly razorDocuments: { [hostDocumentPath: string]: boolean } = {};
+    private readonly razorProjects: { [projectPath: string]: string } = {};
     private readonly documentOpenedEvent = new TelemetryEvent('VSCode.Razor.DocumentOpened');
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Razor.VSCode/src/TelemetryReporter.ts
```diff
@@ -39,6 +41,45 @@ export class TelemetryReporter {
...
+
+        let configurationName: string;
+        let languageVersion: string;
+        if (projectConfiguration.configuration) {
+            configurationName = projectConfiguration.configuration.ConfigurationName;
+            languageVersion = projectConfiguration.configuration.LanguageVersion;
+        } else {
+            configurationName = 'Default';
+            languageVersion = 'Default';
+        }
```
SteveSandersonMS: Don't know if you're into TypeScript tricks, but if you are, lines 51-59 could be replace with:

```typescript
const { ConfigurationName, LanguageVersion }
    = projectConfiguration.configuration || { ConfigurationName: 'Default', LanguageVersion: 'Default' };
```

SteveSandersonMS: Though it is a bit weird that the incoming data is of the shape `project.configuration.configuration.someprops` (with `configuration` repeated). Not saying it's wrong - just a bit quirky!

NTaylorMullen: Not a fan of TypeScript tricks 😄

Issue #193, opened 2018-10-15
Blazor .cshtml files have loads of errors


NTaylorMullen:
#### Is this a Bug or Feature request?:
bug

#### Steps to reproduce
Open a Blazor project.

#### Description of the problem:
You get a ton of errors in Blazor projects a lot like in [this](https://github.com/aspnet/Razor.VSCode/issues/184) issue. Marked this as xs because we can technically treat Blazor projects like Legacy Razor projects until we have actual support.

NTaylorMullen:
f6d717cd1641025bc6402bfd5c8e736a97b0d3c4

PR #204, opened 2018-10-18
Add non-component Blazor support.

NTaylorMullen:
- One of the caveats with this approach is that you end up getting errors on components which are trying to bind event handlers. i.e. `<Counter onclick="@IncrementCount" />`. You get an error stating that you cannot convert a method delegate to object. This is because in non-component scenarios that line is understood as `__o = IncrementCount` where `IncrementCount` is a void returning method.
- Made the experimental language version round trippable once serialized.
- Added the Blazor ProjectEngine to the list of supported VSCode Razor project types.
- Noticed that our testapp projects were incorrectly inheriting our test package references (Xunit and the like).

IF you're viewing a file that uses a component that attaches an event handler via `@SomeMethodHandle` you get the following error:
image

Otherwise, you get Blazor specific C#/directive completion throughout the document like normal i.e. 
image

@danroth27 / @SteveSandersonMS  / @rynowak  you guys need to decide if we should have this non-component Blazor support or force Blazor to have 0 C# completion like a Legacy app, i.e.
image

#193

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.LanguageServer/ProjectEngineFactory_Blazor.cs
+++ b/src/Microsoft.AspNetCore.Razor.LanguageServer/ProjectEngineFactory_Blazor.cs
@@ -1,0 +1,28 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using Microsoft.AspNetCore.Blazor.Razor;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.CodeAnalysis.Razor;
+
+namespace Microsoft.AspNetCore.Razor.LanguageServer
+{
+    // This is a verbatim copy of Blazor's project engine implementation.
+
+    internal class ProjectEngineFactory_Blazor : IProjectEngineFactory
+    {
+        public RazorProjectEngine Create(RazorConfiguration configuration, RazorProjectFileSystem fileSystem, Action<RazorProjectEngineBuilder> configure)
+        {
+            return RazorProjectEngine.Create(configuration, fileSystem, b =>
+            {
+                configure?.Invoke(b);
+                new BlazorExtensionInitializer().Initialize(b);
+
+                var classifier = b.Features.OfType<ComponentDocumentClassifierPass>().Single();
+                classifier.MangleClassNames = true;
+            });
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Razor.LanguageServer/Serialization/RazorConfigurationJsonConverter.cs
+++ b/src/Microsoft.AspNetCore.Razor.LanguageServer/Serialization/RazorConfigurationJsonConverter.cs
@@ -28,10 +28,15 @@ namespace Microsoft.AspNetCore.Razor.LanguageServer.Serialization
 
             var obj = JObject.Load(reader);
             var configurationName = obj[nameof(RazorConfiguration.ConfigurationName)].Value<string>();
-            var languageVersion = obj[nameof(RazorConfiguration.LanguageVersion)].Value<string>();
+            var languageVersionValue = obj[nameof(RazorConfiguration.LanguageVersion)].Value<string>();
             var extensions = obj[nameof(RazorConfiguration.Extensions)].ToObject<RazorExtension[]>(serializer);
 
-            return RazorConfiguration.Create(RazorLanguageVersion.Parse(languageVersion), configurationName, extensions);
+            if (!RazorLanguageVersion.TryParse(languageVersionValue, out var languageVersion))
+            {
+                languageVersion = RazorLanguageVersion.Version_2_1;
+            }
+
+            return RazorConfiguration.Create(languageVersion, configurationName, extensions);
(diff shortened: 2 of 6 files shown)


src/Microsoft.AspNetCore.Razor.LanguageServer/Serialization/RazorConfigurationJsonConverter.cs
```diff
@@ -28,10 +28,15 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
 
             var obj = JObject.Load(reader);
             var configurationName = obj[nameof(RazorConfiguration.ConfigurationName)].Value<string>();
-            var languageVersion = obj[nameof(RazorConfiguration.LanguageVersion)].Value<string>();
+            var languageVersionValue = obj[nameof(RazorConfiguration.LanguageVersion)].Value<string>();
             var extensions = obj[nameof(RazorConfiguration.Extensions)].ToObject<RazorExtension[]>(serializer);
 
-            return RazorConfiguration.Create(RazorLanguageVersion.Parse(languageVersion), configurationName, extensions);
+            if (!RazorLanguageVersion.TryParse(languageVersionValue, out var languageVersion))
```
NTaylorMullen: @rynowak I found this kind of interesting that our experimental language version wasn't round-trippable. This wasn't intentional was it?

src/Microsoft.AspNetCore.Razor.LanguageServer/ProjectEngineFactory_Blazor.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Linq;
+using Microsoft.AspNetCore.Blazor.Razor;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.CodeAnalysis.Razor;
+
+namespace Microsoft.AspNetCore.Razor.LanguageServer
+{
+    // This is a verbatim copy of Blazor's project engine implementation.
+
+    internal class ProjectEngineFactory_Blazor : IProjectEngineFactory
```
rynowak: Why this wierd naming instead of `BlazorProjectEngineFactory`?

src/Microsoft.AspNetCore.Razor.LanguageServer/Serialization/RazorConfigurationJsonConverter.cs
rynowak: No real idea, but I'm sure you'd be able to fix it.

src/Microsoft.AspNetCore.Razor.LanguageServer/Serialization/RazorConfigurationJsonConverter.cs
```diff
@@ -28,10 +28,15 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
...
+            {
+                languageVersion = RazorLanguageVersion.Version_2_1;
+            }
+
+            return RazorConfiguration.Create(languageVersion, configurationName, extensions);
```
rynowak: I'm really concerned about the fact that you took a non-shared source version of this code, and you're now maintiaining it. What's the plan to unify this?

src/Microsoft.AspNetCore.Razor.LanguageServer/Serialization/RazorConfigurationJsonConverter.cs
rynowak: Is the issue here that experimental doesn't parse?

NTaylorMullen: Correct. It tries to parse `1337.1337` and fails because it can only parse `experimental`

src/Microsoft.AspNetCore.Razor.LanguageServer/ProjectEngineFactory_Blazor.cs
NTaylorMullen: Consistency

src/Microsoft.AspNetCore.Razor.LanguageServer/Serialization/RazorConfigurationJsonConverter.cs
NTaylorMullen: This is more of an artifact of being in the "alpha" phase of development. For a while we couldn't depend on Razor packages because of strong naming restrictions. Now that that is resolved I just need to go back through the code and change it all to use the proper packages. 

In this specific case I couldn't find a package for the `Microsoft.VisualStudio.LanguageServices.Blazor` so I hacked together a solution since i'll have to go back and hot swap everything else anyhow.

NTaylorMullen: Oh and we also need to change the tooling packages/projects to cross-compile for something netcoreapp2.0 compatible

danroth27:
> you guys need to decide if we should have this non-component Blazor support or force Blazor to have 0 C# completion like a Legacy app

I think the Blazor support is compelling enough that I prefer to ship it even with the errors showing up. I can add a Known Issue to let folks know the error can be ignored.

PR #212, opened 2018-10-25
Add Client.csproj. Fixes #139

### LINKED ISSUE #139: Running ./build doesn't npm install the dev client.

SteveSandersonMS:


SteveSandersonMS:
Not merging this right now because the build is *sometimes* failing on `Razor.VSCode-CI-Public` even though it always works on `Razor.VSCode-CI-Official`. The intermittent failure is this error:

```
npm ERR! enoent ENOENT: no such file or directory, rename 'D:\a\1\s\client\node_modules\.staging\microsoft.aspnetcore.razor.vscode-90813625\node_modules\@types\node' -> 'D:\a\1\s\client\node_modules\.staging\@types\node-ea9b1830'
```

This is a familiar problem. NPM has a long history of intermittent failures of this kind on Windows. Usually they are fixed by upgrading or downgrading to a different version of NPM. Since `Razor.VSCode-CI-Official` uses a different version of NPM (6.0.1, where it works) than is used on `Razor.VSCode-CI-Public` (6.4.1, where it doesn't work), that most likely explains it.

@natemcmaster Do you know if there's a way to configure which version of Node/NPM should be installed in the build environment?

natemcmaster:
Yes, you can configure the node version with this build task: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/node-js

SteveSandersonMS:
Thanks @natemcmaster - that looks perfect! The docs there about all the tasks look really useful too.

SteveSandersonMS:
OK, should work reliably now.

PR #213, opened 2018-10-25
Factor our common parts of build config into template

SteveSandersonMS:


SteveSandersonMS:
This is producing correct builds now, so merging.

PR #225, opened 2018-11-09
Updated dependencies.props and korebuild information.

NTaylorMullen:
- With the new korebuild we can no longer rely on specialized project references to enforce omnisharp plugin building. Instead, we now force build the plugin manually. This in turn has resulted in a faster build.


src/Microsoft.AspNetCore.Razor.LanguageServer/Microsoft.AspNetCore.Razor.LanguageServer.csproj
```diff
@@ -19,17 +19,14 @@
...
-    <!-- We need this to compile but don't want it referenced by the project. -->
-    <ProjectReference Include="..\Microsoft.AspNetCore.Razor.OmniSharpPlugin\Microsoft.AspNetCore.Razor.OmniSharpPlugin.csproj">
-      <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
-      <Targets>Publish</Targets>
-    </ProjectReference>
-  </ItemGroup>
-
   <Target Name="PublishLanguageServerNativeExecutables" AfterTargets="Pack" DependsOnTargets="PublishAllRids" />
 
   <Target Name="IncludeOmniSharpPlugin" AfterTargets="Publish" Condition="Exists('$(PublishDir)')">
```
SteveSandersonMS: Does this really only need to happen on `Publish`? Isn't it also needed during development? Or are you assuming that developers always start by running `build.cmd` which implictly does a publish (not just a build)?

src/Microsoft.AspNetCore.Razor.LanguageServer/Microsoft.AspNetCore.Razor.LanguageServer.csproj
```diff
@@ -19,17 +19,14 @@
...
+      <MSBuild
+        Projects="..\Microsoft.AspNetCore.Razor.OmniSharpPlugin\Microsoft.AspNetCore.Razor.OmniSharpPlugin.csproj"
+        Properties="PublishDir=$(MSBuildProjectDirectory)\..\Microsoft.AspNetCore.Razor.OmniSharpPlugin\bin\$(Configuration)\net461\publish;RuntimeIdentifier="
```
SteveSandersonMS: Not sure why this specifies `RuntimeIdentifier=`, but that's fine if there is some reason for it!

src/Microsoft.AspNetCore.Razor.LanguageServer/Microsoft.AspNetCore.Razor.LanguageServer.csproj
```diff
@@ -19,17 +19,14 @@
...
+        Targets="Publish" />
```
SteveSandersonMS: Nit: Indentation in this new block looks off.

Issue #210, opened 2018-10-22
Incorrect completion in a simple view page


akshita31:
#### Is this a Bug or Feature request?:
Bug

#### Steps to reproduce
Open a simple ASP.NET application like the one [here](https://github.com/akshita31/ASP.NET-Learning-Sample/blob/v1.0/WebApplication1/Views/Home/Index.cshtml#L1) the project here and type @model SampleApplication.Models.

#### Description of the problem:
Unexpected completion when typing Models. in continuation. When typing just "." after Models then the expected completion is shown.

Expected behavior:
Only Restaurants is shown in the completion

Actual behavior:
"models" is shown in completions which is unexpected
![razor](https://user-images.githubusercontent.com/12554141/47282806-7552bc80-d595-11e8-91ee-8bd8a5271a88.gif)

#### Additional information

1. Razor C# output
[razorCSharp.txt](https://github.com/aspnet/Razor.VSCode/files/2500544/razorCSharp.txt)

2. Razor log Output
[razorLog.txt](https://github.com/aspnet/Razor.VSCode/files/2500527/razorLog.txt)

4. OmniSharp log output
[omnisharpLog.txt](https://github.com/aspnet/Razor.VSCode/files/2500533/omnisharpLog.txt)

NTaylorMullen:
Seeing this more frequently now, diagnosing it a little bit it looks like an issue on the OmniSharp side. Need to dive a little deeper but basically OmniSharp is returning completion info for an outdated document.

akshita31:
@NTaylorMullen It just shows the plural of whatever preceeds the dot. I randomly tried SampleApplication.Services. and it showed "services" as the suggestion

NTaylorMullen:
> @NTaylorMullen It just shows the plural of whatever preceeds the dot. I randomly tried SampleApplication.Services. and it showed "services" as the suggestion

You mean the lowercase variant?

akshita31:
Lowercase variant as well as plural. For "abc", I am getting "abcs"

NTaylorMullen:
> Lowercase variant as well as plural. For "abc", I am getting "abcs"

Ahhh, good to know. Must be another completion provider in VSCode providing that. Good catch @akshita31 !

NTaylorMullen:
[...]

PR #226, opened 2018-11-09
Re-design document synchronization.

NTaylorMullen:
- The purpose of this redesign is to make completion more reliable. We now have a great understanding of how VSCode's object model mutates and therefore can make better assumptiosn on how to synchronize it.
- In this changeset we primarily do 2 things to synchronize a document:
  1. Make sure our `IProjectedDocument` has been updated by our language server.
  2. Make sure VSCode's TextDocument reflects the appropriate content detailed by the `IProjectedDocument`
If these two assertions are true then every other extension sees updated C#/Html and can respond accordingly.
- To properly accomplish 2 I had to persist information into the `vscode.TextDocument` to identify when the content properly matches our DTO. We do this by adding a `// SomeNumber` to the end of projected text documents.
- Updated RazorDocumentFactory to generate virtual C# files without a `__` prefix to prevent URIs from being placed as "authoritative" sources labeled by `__`. This caused issues in VSCode when working with copmletion.
- Flowed cancellation tokens into synchronization APIs so that when VSCode cancels an action we can also cancel synchronization.

#210

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.VSCode/src/Html/HtmlProjectedDocumentContentProvider.ts
+++ b/src/Microsoft.AspNetCore.Razor.VSCode/src/Html/HtmlProjectedDocumentContentProvider.ts
@@ -27,7 +27,9 @@ export class HtmlProjectedDocumentContentProvider implements vscode.TextDocument
             return;
         }
 
-        const content = razorDocument.htmlDocument.getContent();
+        const content = `${razorDocument.htmlDocument.getContent()}
+// ${razorDocument.htmlDocument.projectedDocumentSyncVersion}`;
+
         return content;
     }
 
--- a/src/Microsoft.AspNetCore.Razor.VSCode/src/IProjectedDocument.ts
+++ b/src/Microsoft.AspNetCore.Razor.VSCode/src/IProjectedDocument.ts
@@ -9,5 +9,6 @@ export interface IProjectedDocument {
     readonly path: string;
     readonly uri: vscode.Uri;
     readonly hostDocumentSyncVersion: number | null;
+    readonly projectedDocumentSyncVersion: number;
     getContent(): string;
 }
--- a/src/Microsoft.AspNetCore.Razor.VSCode/src/RazorDocumentFactory.ts
+++ b/src/Microsoft.AspNetCore.Razor.VSCode/src/RazorDocumentFactory.ts
@@ -27,8 +27,8 @@ export function createDocument(uri: vscode.Uri) {
 }
 
 function createProjectedHtmlDocument(hostDocumentUri: vscode.Uri) {
-    // Index.cshtml => __Index.cshtml.html
-    const projectedPath = `__${hostDocumentUri.path}.html`;
+    // Index.cshtml => Index.cshtml__virtual.html
+    const projectedPath = `${hostDocumentUri.path}__virtual.html`;
     const projectedUri = vscode.Uri.parse(`${HtmlProjectedDocumentContentProvider.scheme}://${projectedPath}`);
     const projectedDocument = new HtmlProjectedDocument(projectedUri);
 
@@ -36,8 +36,8 @@ function createProjectedHtmlDocument(hostDocumentUri: vscode.Uri) {
 }
 
 function createProjectedCSharpDocument(hostDocumentUri: vscode.Uri) {
-    // Index.cshtml => __Index.cshtml__virtual.cs
-    const projectedPath = `__${hostDocumentUri.path}__virtual.cs`;
+    // Index.cshtml => Index.cshtml__virtual.cs
+    const projectedPath = `${hostDocumentUri.path}__virtual.cs`;
     const projectedUri = vscode.Uri.parse(`${CSharpProjectedDocumentContentProvider.scheme}://${projectedPath}`);
     const projectedDocument = new CSharpProjectedDocument(projectedUri);
 
--- a/src/Microsoft.AspNetCore.Razor.VSCode/src/RazorDocumentSynchronizer.ts
+++ b/src/Microsoft.AspNetCore.Razor.VSCode/src/RazorDocumentSynchronizer.ts
@@ -5,183 +5,271 @@
 
 import * as vscode from 'vscode';
 
+import { CSharpProjectedDocumentContentProvider } from './CSharp/CSharpProjectedDocumentContentProvider';
+import { HtmlProjectedDocumentContentProvider } from './Html/HtmlProjectedDocumentContentProvider';
 import { IProjectedDocument } from './IProjectedDocument';
 import { IRazorDocumentChangeEvent } from './IRazorDocumentChangeEvent';
 import { RazorDocumentChangeKind } from './RazorDocumentChangeKind';
 import { RazorDocumentManager } from './RazorDocumentManager';
-import { RazorLanguage } from './RazorLanguage';
 import { RazorLogger } from './RazorLogger';
 import { getUriPath } from './UriPaths';
 
 export class RazorDocumentSynchronizer {
-    private readonly synchronizations: { [uri: string]: SynchronizationContext } = {};
+    private readonly synchronizations: { [uri: string]: SynchronizationContext[] } = {};
     private synchronizationIdentifier = 0;
 
     constructor(
-        documentManager: RazorDocumentManager,
+        private readonly documentManager: RazorDocumentManager,
         private readonly logger: RazorLogger) {
-        documentManager.onChange((event) => this.documentChanged(event));
     }
 
     public register() {
-        const changeRegistration = vscode.workspace.onDidChangeTextDocument((args) => {
-            if (args.document.languageId !== RazorLanguage.id) {
-                return;
-            }
+        const documentManagerRegistration = this.documentManager.onChange(
+            event => this.documentChanged(event));
+        const textDocumentChangeRegistration = vscode.workspace.onDidChangeTextDocument(
+            event => this.textDocumentChanged(event));
 
-            const uriPath = getUriPath(args.document.uri);
-            const context = this.synchronizations[uriPath];
-
-            if (context && args.document.version >= context.documentVersion) {
-                if (this.logger.verboseEnabled) {
-                    this.logger.logVerbose(
-                        `${context.logIdentifier} - Notify Success: Host document updated to equivalent version.`);
-                }
-                context.synchronized(true);
-            }
-        });
-
-        return changeRegistration;
+        return vscode.Disposable.from(documentManagerRegistration, textDocumentChangeRegistration);
     }
 
-    public async trySynchronize(
+    public async trySynchronizeProjectedDocument(
         hostDocument: vscode.TextDocument,
         projectedDocument: IProjectedDocument,
-        toVersion: number) {
-        const logIdentifier = this.synchronizationIdentifier++;
+        expectedHostDocumentVersion: number,
+        token: vscode.CancellationToken) {
+
+        const logId = ++this.synchronizationIdentifier;
+
+        const documentKey = getUriPath(projectedDocument.uri);
+        if (this.logger.verboseEnabled) {
+            const ehdv = expectedHostDocumentVersion;
+            this.logger.logVerbose(
+                `${logId} - Synchronizing '${documentKey}':
+    Currently at ${projectedDocument.hostDocumentSyncVersion}, synchronizing to version '${ehdv}'.
+    Current host document version: '${hostDocument.version}'
+    Current projected document version: '${projectedDocument.projectedDocumentSyncVersion}'`);
+        }
+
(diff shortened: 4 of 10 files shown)


src/Microsoft.AspNetCore.Razor.VSCode/src/Html/HtmlProjectedDocumentContentProvider.ts
```diff
@@ -27,7 +27,9 @@ export class HtmlProjectedDocumentContentProvider implements vscode.TextDocument
             return;
         }
 
-        const content = razorDocument.htmlDocument.getContent();
+        const content = `${razorDocument.htmlDocument.getContent()}
+// ${razorDocument.htmlDocument.projectedDocumentSyncVersion}`;
```
SteveSandersonMS: This is a bit quirky, seeing as `//` is not the syntax for comments in HTML. However I can't think of any problem that it's going to cause (since we only use the projected HTML doc for completions, etc., and the user isn't going to actually see the projected document rendered anywhere).

You could even eliminate the `//` if you wanted, as the effect would be the same.

src/Microsoft.AspNetCore.Razor.VSCode/src/IProjectedDocument.ts
```diff
@@ -9,5 +9,6 @@ export interface IProjectedDocument {
     readonly path: string;
     readonly uri: vscode.Uri;
     readonly hostDocumentSyncVersion: number | null;
+    readonly projectedDocumentSyncVersion: number | null;
```
SteveSandersonMS: How can this be `null`? It starts at `0`.

src/Microsoft.AspNetCore.Razor.VSCode/src/RazorDocumentFactory.ts
```diff
@@ -27,17 +27,17 @@ export function createDocument(uri: vscode.Uri) {
 }
 
 function createProjectedHtmlDocument(hostDocumentUri: vscode.Uri) {
-    // Index.cshtml => __Index.cshtml.html
-    const projectedPath = `__${hostDocumentUri.path}.html`;
+    // Index.cshtml => Index.cshtml__virtual.html
+    const projectedPath = `${hostDocumentUri.path}__virtual.html`;
```
SteveSandersonMS: TBH I haven't fully understood the scenarios where this path shows up anywhere that the change would make a difference. Can the user ever see the `*__virtual.html` filenames anywhere?

src/Microsoft.AspNetCore.Razor.VSCode/src/RazorDocumentSynchronizer.ts
```diff
@@ -24,164 +25,257 @@ export class RazorDocumentSynchronizer {
...
+        // Logic defined in this method is heavily dependent on the functionality in the projected
+        // document content providers to append versions to the end of text documents.
 
+        if (textDocument.lineCount <= 0) {
+            return null;
         }
+
+        const lastLine = textDocument.lineAt(textDocument.lineCount - 1);
+
+        if (!lastLine.text.startsWith('// ')) {
```
SteveSandersonMS: Not sure I buy this logic.

1. If there's a possibility that the last line might *not* start with `//`, then that implies the system doesn't always insert the `// (number)` line.
2. As a consequence, it's possible that the last line was something the developer wrote themselves.
3. The thing the developer wrote themselves might start with `// ` but not be followed by a number.
4. In this case, we'll get a `NaN` on line 264, or possibly even a random other number. Not sure how to reason about what the effect of this would be.

If we have to distinguish between system-added lines and developer-written ones, could we at least make accidental conflicts unlikely by changing the pattern to something like `// [VS Code Razor sync version {number}]`?

src/Microsoft.AspNetCore.Razor.VSCode/src/RazorDocumentSynchronizer.ts
```diff
@@ -24,164 +25,257 @@ export class RazorDocumentSynchronizer {
     }
```
SteveSandersonMS: On the preceding line, it seems odd that we register for events both in the constructor and in a public `register` method. Would it make sense to put both registrations in the same place (either the constructor, or in `register`)? Or is the distinction between the two necessary for some functionality?

src/Microsoft.AspNetCore.Razor.VSCode/src/RazorDocumentSynchronizer.ts
```diff
@@ -24,164 +25,257 @@ export class RazorDocumentSynchronizer {
...
-                `Current host document version: '${hostDocument.version}'`);
+            this.logger.logVerbose(
+                `${logId} - Synchronizing '${documentKey}':
+    Currently at ${projectedDocument.hostDocumentSyncVersion}, synchronizing to version '${toHostDocumentVersion}'.
+    Current host document version: '${hostDocument.version}'
+    Current projected document version: '${projectedDocument.projectedDocumentSyncVersion}'`);
         }
 
-        if (projectedDocument.hostDocumentSyncVersion === hostDocument.version) {
+        if (hostDocument.version !== toHostDocumentVersion) {
```
SteveSandersonMS: This check feels odd to me. Maybe I'm just misreading it - hope you can clarify.

The caller has asked for `hostDocument` to be synchronized to version `toHostDocumentVersion`. So the caller probably doesn't expect for `hostDocument`'s version already to equal `toHostDocumentVersion` (otherwise they wouldn't be calling this). So why do we log a warning to say the version doesn't already match?

Maybe some renaming of args would clarify the intent.

SteveSandersonMS: Oh wait I think I understand now. This method intends to synchronize the *projected* document with the host document. Your check is to validate that the caller is correct about the current version of the host document, not the projected document that we are updating. I don't 100% know why we need to check that (instead of dropping the `toHostDocumentVersion` param, and just always updating the projected document to match `hostDocument`) but it's fine if you have some reason!

Maybe to clarify, this method could be renamed to `trySynchronizeProjectedDocument`, and the `toHostDocumentVersion` param could be renamed to `expectedHostDocumentVersion`.

src/Microsoft.AspNetCore.Razor.VSCode/src/RazorDocumentSynchronizer.ts
```diff
@@ -24,164 +25,257 @@ export class RazorDocumentSynchronizer {
...
+            return;
+        }
+
+        for (let i = 0; i < synchronizations.length; i++) {
+            if (synchronizations[i] === context) {
+                const newSynchronizations = synchronizations.splice(i, 1);
+                this.synchronizations[documentKey] = newSynchronizations;
+                break;
             }
+        }
```
SteveSandersonMS: Nitpick: Lines 129-134 could be:

```javascript
this.synchronizations[documentKey] = synchronizations.filter(item => item !== context);
```

src/Microsoft.AspNetCore.Razor.VSCode/src/RazorDocumentSynchronizer.ts
```diff
@@ -24,164 +25,257 @@ export class RazorDocumentSynchronizer {
...
-            return false;
+    private createSynchronizationContext(
+        documentKey: string,
+        projectedDocument: IProjectedDocument,
+        toHostDocumentVersion: number,
+        hostDocument: vscode.TextDocument,
+        token: vscode.CancellationToken) {
+
+        const rejectionsForCancel: Array<(reason: string) => void> = [];
+        let projectedDocumentSynchronized: () => void = Function;
```
SteveSandersonMS: This is interesting. I've never seen the `Function` object used as an actual function instance before.

Is this definitely what you want? I just tried it, and at least in browsers, attempting to invoke `Function()` triggers a CSP exception. Maybe it's OK in VS Code's JS runtime. Perhaps it would be more conventional to replace this with:

```javascript
let projectedDocumentSynchronized = () => {};
```

... if you definitely need a no-op. I know that will allocate more but so is everything else.

SteveSandersonMS: Oh it looks like you're not using the initial value anyway. So it could be:

```javascript
let projectedDocumentSynchronized: Function;
```

src/Microsoft.AspNetCore.Razor.VSCode/src/Html/HtmlProjectedDocumentContentProvider.ts
NTaylorMullen: > You could even eliminate the `//` if you wanted, as the effect would be the same.

Can't because it'd result in the C# document being invalid. I ould technically do something different for Html but there's no harm in having `//` so for the sake of consistency I did the same for both virtual document.

src/Microsoft.AspNetCore.Razor.VSCode/src/RazorDocumentFactory.ts
NTaylorMullen: No they cannot, this was a fix I added after debugging through vscode itself. In their internal representation of our files they strip the `__` and have their URI authority as `__`. Therefore, when they try and cross-lookup our files they request funky paths with `__` at the beginning of the rooted path.

src/Microsoft.AspNetCore.Razor.VSCode/src/RazorDocumentSynchronizer.ts
NTaylorMullen: > 1. If there's a possibility that the last line might _not_ start with `//`, then that implies the system doesn't always insert the `// (number)` line.

Not possible because we own the document generation. This is only looking at virtual documents that we serve (and ensure that the last line is `// (number)`



> 2\. As a consequence, it's possible that the last line was something the developer wrote themselves.

Can't because we append a line after the last line that the developer wrote.

> 3\. The thing the developer wrote themselves might start with `// ` but not be followed by a number.

That's fine, it wont be the last line.

> 4\. In this case, we'll get a `NaN` on line 264, or possibly even a random other number. Not sure how to reason about what the effect of this would be.

Because of the above (don't need to distinguish) we don't need to worry about this.

src/Microsoft.AspNetCore.Razor.VSCode/src/RazorDocumentSynchronizer.ts
NTaylorMullen: Ya I can definitely do that

src/Microsoft.AspNetCore.Razor.VSCode/src/RazorDocumentSynchronizer.ts
NTaylorMullen: Hmmm, love the new naming (and you're dead on for what it's trying to accomplish), i'll definitely do that.

src/Microsoft.AspNetCore.Razor.VSCode/src/RazorDocumentSynchronizer.ts
SteveSandersonMS: So you don't need line 259 then?

src/Microsoft.AspNetCore.Razor.VSCode/src/RazorDocumentSynchronizer.ts
NTaylorMullen: Btw, how bad would it be if we nuked the line length restriction in TypeScript? There are so many places where I feel like the restriction is more harmful then beneficial

src/Microsoft.AspNetCore.Razor.VSCode/src/RazorDocumentSynchronizer.ts
NTaylorMullen: Fair, yup don't really need that 😄

src/Microsoft.AspNetCore.Razor.VSCode/src/RazorDocumentSynchronizer.ts
NTaylorMullen: Have to assign it to something because:
image

And tslint complains about using `Function` as a type:
image

Either way the variable immediately gets set in the promise creation. I was doing this with the intent of pre-setting a variable to `null` like in C#. There a better way to work around the limitations in  clean way?

SteveSandersonMS: Setting it to `null` would be fine (`null` !== `Function`).

NTaylorMullen: Would have to null check it before invoking it then :/. Gonna leave it as Function for now, it'll be immediately overridden anyhow.

Issue #222, opened 2018-11-02
Consider: Don't start razor language service when there are no razor files


gregg-miskelly:
_From @gregg-miskelly on November 2, 2018 22:51_

## Environment data
C# Extension version: 1.17

## Steps to reproduce
1. Create a new C# .NET Core console app
2. Open the folder in VS Code

## Expected  behavior
Razor language service (rzls) process shouldn't start since there are no razor files

## Actual behavior
Razor language service starts

_Copied from original issue: OmniSharp/omnisharp-vscode#2649_

gregg-miskelly:
_From @NTaylorMullen on November 2, 2018 22:54_

This is definitely a reasonable request. @gregg-miskelly would you mind filing the issue over in https://github.com/aspnet/Razor.VSCode?

gregg-miskelly:
I can move it using https://github-issue-mover.appspot.com

NTaylorMullen:
86b17fa24f45ae0cddbce9b93f3bc583150d1b0d

PR #245, opened 2018-11-27
Don't start language server if there are no Razor files in the workspace.

NTaylorMullen:
- Updated dev package.json to reflect omnisharp activation events.
- We delay the start of the language server until a Razor file has been created or a Razor file has been opened. One thing to note, is when opening a Razor file in an unrelated workspace you run into #244.
- Added an ensure mechanism to the `RazorLanguageServiceClient` to ensure that any requests that need to make their way through the system will wait on the server starting since we now delay its start.
- Added an `alwaysLog`method to the logger to allow us to notify users when we're delay starting the language server.
- Had to add an `onStarted` event to the server client in order to allow for initialization of our document and project managers.

#222

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.VSCode/src/RazorLanguageServerClient.ts
+++ b/src/Microsoft.AspNetCore.Razor.VSCode/src/RazorLanguageServerClient.ts
@@ -26,8 +26,10 @@ export class RazorLanguageServerClient implements vscode.Disposable {
     private serverOptions: ServerOptions;
     private client: LanguageClient;
     private startDisposable: vscode.Disposable | undefined;
+    private onStartedListeners: Array<() => Promise<any>> = [];
     private eventBus: EventEmitter;
     private isStarted: boolean;
+    private startHandle: Promise<void> | undefined;
 
     constructor(
         options: RazorLanguageServerOptions,
@@ -73,6 +75,10 @@ export class RazorLanguageServerClient implements vscode.Disposable {
         this.eventBus = new EventEmitter();
     }
 
+    public onStarted(listener: () => Promise<any>) {
+        this.onStartedListeners.push(listener);
+    }
+
     public onStart(listener: () => any) {
         this.eventBus.addListener(events.ServerStart, listener);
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNetCore.Razor.VSCode/src/RazorLanguageServerClient.ts
```diff
@@ -73,6 +75,10 @@ export class RazorLanguageServerClient implements vscode.Disposable {
         this.eventBus = new EventEmitter();
     }
 
+    public onStarted(listener: () => Promise<any>) {
+        this.onStartedListeners.push(listener);
```
SteveSandersonMS: Is there any scenario where someone could try to register an `onStarted` listener after it's already started? If so, what's the desired behavior?

 * We could call their listener callback immediately
 * We could ignore them

The code as-is would accumulate extra listeners in `onStartedListeners` but never do anything with them. This is fine if we have reason to believe the scenario will never occur.

src/Microsoft.AspNetCore.Razor.VSCode/src/RazorLanguageServerClient.ts
```diff
@@ -90,22 +96,42 @@ export class RazorLanguageServerClient implements vscode.Disposable {
...
-            this.logger.logMessage('Server started and ready!');
 
             this.isStarted = true;
+            this.logger.logMessage('Server started and ready!');
             this.eventBus.emit(events.ServerStart);
+
+            for (const listener of this.onStartedListeners) {
+                await listener();
+            }
+
```
SteveSandersonMS: Not essential, but we *could* empty `this.onStartedListeners` here so we're not preventing them and their chain of references from being GCed.

src/Microsoft.AspNetCore.Razor.VSCode/src/extension.ts
```diff
@@ -89,12 +89,43 @@ export async function activate(context: ExtensionContext, languageServerDir: str
...
+        const watcher = vscode.workspace.createFileSystemWatcher(RazorLanguage.globbingPattern);
+        const delayedLanguageServerStart = async () => {
+            razorFileCreatedRegistration.dispose();
+            razorFileOpenedRegistration.dispose();
+            await languageServerClient.start();
+        };
+        const razorFileCreatedRegistration = watcher.onDidCreate(() => delayedLanguageServerStart());
+        const razorFileOpenedRegistration = vscode.workspace.onDidOpenTextDocument(async (event) => {
+            if (event.languageId === RazorLanguage.id) {
+                await delayedLanguageServerStart();
```
SteveSandersonMS: Not sure if there's a reason we `await delayedLanguageServerStart` here but don't `await` it in the `onDidCreate` callback.

Should line 121 be changed to the following for consistency?

```typescript
const razorFileCreatedRegistration = watcher.onDidCreate(delayedLanguageServerStart);
```

... or is the reason that the `onDidOpenTextDocument` event expects async listeners whereas `onDidCreate` expects synchronous ones?

src/Microsoft.AspNetCore.Razor.VSCode/src/RazorLanguageServerClient.ts
NTaylorMullen: Scenario will never occurr

src/Microsoft.AspNetCore.Razor.VSCode/src/extension.ts
NTaylorMullen: the onDidCreate callback doesn't allow awaiting anything sadly but I figured i'd still use the `() => ` syntax to ensure that if we ever change the `delayedLanguageServerSTart` logic to depend on `this` we don't break things.

src/Microsoft.AspNetCore.Razor.VSCode/src/RazorLanguageServerClient.ts
NTaylorMullen: Eh, they shouldn't be super significant and disposing of the server client object should be sufficient enough to allow for GC.

Issue #182, opened 2018-10-08
Add issue reporter command


NTaylorMullen:
It would be equivalent to what's going on here: https://github.com/OmniSharp/omnisharp-vscode/pull/2592

The general idea is to make filing issues easier when it comes to reporting issues for Razor VSCode.

PR #259, opened 2018-12-18
Report a Razor issue feature.

NTaylorMullen:
From a design perspective I'm looking for feedback on how this works. Had to workaround several limitations in VSCode and this is what I came up with. Working on building unit tests for this feature.

- Added the ability to programatically collect usage data in order to make filing Razor issues easier.
- Built a report issue panel that can be interacted with to control how data is collected.
- There wasn't any valid APIs to navigate to GitHub.com's create new issue location with a large payload as the body. Therefore, I took the approach of copying issue content to the clipboard in order to aid in issue creation.
- Expanded the line length limit in tslint. 120 was too restricting and led to hard-to-read code.

How it looks (in two gifs because of upload limitations):
image
image

The full issue content: 
[issueContent.txt](https://github.com/aspnet/Razor.VSCode/files/2691935/issueContent.txt)

#182 

FYI @akshita31 @rchande @danroth27

### DIFF
--- a/client/unittest/ReportIssueCreator.test.ts
+++ b/client/unittest/ReportIssueCreator.test.ts
@@ -1,0 +1,291 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+
+import * as assert from 'assert';
+import { IReportIssueDataCollectionResult } from 'microsoft.aspnetcore.razor.vscode/dist/Diagnostics/IReportIssueDataCollectionResult';
+import { ReportIssueCreator } from 'microsoft.aspnetcore.razor.vscode/dist/Diagnostics/ReportIssueCreator';
+import { IRazorDocument } from 'microsoft.aspnetcore.razor.vscode/dist/IRazorDocument';
+import { IRazorDocumentManager } from 'microsoft.aspnetcore.razor.vscode/dist/IRazorDocumentManager';
+import * as vscode from 'microsoft.aspnetcore.razor.vscode/dist/vscodeAdapter';
+import { TestProjectedDocument } from './Mocks/TestProjectedDocument';
+import { TestRazorDocument } from './Mocks/TestRazorDocument';
+import { TestRazorDocumentManager } from './Mocks/TestRazorDocumentManager';
+import { TestTextDocument } from './Mocks/TestTextDocument';
+import { createTestVSCodeApi } from './Mocks/TestVSCodeApi';
+
+describe('ReportIssueCreator', () => {
+    function getReportIssueCreator(api: vscode.api) {
+        const documentManager = new TestRazorDocumentManager();
+        const issueCreator = new TestReportIssueCreator(api, documentManager);
+        return issueCreator;
+    }
+
+    it('create can operate when no content is available', async () => {
+        // Arrange
+        const api = createTestVSCodeApi();
+        const issueCreator = getReportIssueCreator(api);
+        const collectionResult: IReportIssueDataCollectionResult = {
+            document: undefined,
+            logOutput: '',
+        };
+
+        // Act
+        const issueContent = await issueCreator.create(collectionResult);
+
+        // Assert
+        assert.ok(issueContent.indexOf('Bug') > 0);
+    });
+
+    it('getRazor returns text documents contents', async () => {
+        // Arrange
+        const api = createTestVSCodeApi();
+        const issueCreator = getReportIssueCreator(api);
+        const expectedContent = 'TextDocument content';
+        const textDocument = new TestTextDocument(expectedContent, api.Uri.parse('C:/path/to/file.cshtml'));
+
+        // Act
+        const razorContent = await issueCreator.getRazor(textDocument);
+
+        // Assert
+        assert.equal(razorContent, expectedContent);
+    });
+
+    it('getProjectedCSharp returns projected CSharp and vscodes text document CSharp', async () => {
+        // Arrange
+        const api = createTestVSCodeApi();
+        const expectedVSCodeCSharpContent = 'VSCode seen CSharp content';
+        const csharpDocumentUri = api.Uri.parse('C:/path/to/file.cshtml.__virtual.cs');
+        const csharpTextDocument = new TestTextDocument(expectedVSCodeCSharpContent, csharpDocumentUri);
+        api.setWorkspaceDocuments(csharpTextDocument);
+        const hostDocumentUri = api.Uri.parse('C:/path/to/file.cshtml');
+        const expectedProjectedCSharpContent = 'Projected CSharp content';
+        const razorDocument = new TestRazorDocument(hostDocumentUri, hostDocumentUri.path);
+        razorDocument.csharpDocument = new TestProjectedDocument(expectedProjectedCSharpContent, csharpDocumentUri);
+        const issueCreator = getReportIssueCreator(api);
+
+        // Act
+        const razorContent = await issueCreator.getProjectedCSharp(razorDocument);
+
+        // Assert
+        assert.ok(razorContent.indexOf(expectedVSCodeCSharpContent) > 0);
+        assert.ok(razorContent.indexOf(expectedProjectedCSharpContent) > 0);
+    });
+
+    it('getProjectedCSharp returns only projected CSharp if cannot locate vscodes text document', async () => {
+        // Arrange
+        const api = createTestVSCodeApi();
+        const csharpDocumentUri = api.Uri.parse('C:/path/to/file.cshtml.__virtual.cs');
+        const hostDocumentUri = api.Uri.parse('C:/path/to/file.cshtml');
+        const expectedProjectedCSharpContent = 'Projected CSharp content';
+        const razorDocument = new TestRazorDocument(hostDocumentUri, hostDocumentUri.path);
+        razorDocument.csharpDocument = new TestProjectedDocument(expectedProjectedCSharpContent, csharpDocumentUri);
+        const issueCreator = getReportIssueCreator(api);
+
+        // Act
+        const razorContent = await issueCreator.getProjectedCSharp(razorDocument);
+
+        // Assert
+        assert.ok(razorContent.indexOf(expectedProjectedCSharpContent) > 0);
+    });
+
+    it('getProjectedHtml returns projected Html and vscodes text document Html', async () => {
+        // Arrange
+        const api = createTestVSCodeApi();
+        const expectedVSCodeHtmlContent = 'VSCode seen Html content';
+        const htmlDocumentUri = api.Uri.parse('C:/path/to/file.cshtml.__virtual.cs');
+        const htmlTextDocument = new TestTextDocument(expectedVSCodeHtmlContent, htmlDocumentUri);
+        api.setWorkspaceDocuments(htmlTextDocument);
+        const hostDocumentUri = api.Uri.parse('C:/path/to/file.cshtml');
+        const expectedProjectedHtmlContent = 'Projected Html content';
+        const razorDocument = new TestRazorDocument(hostDocumentUri, hostDocumentUri.path);
+        razorDocument.htmlDocument = new TestProjectedDocument(expectedProjectedHtmlContent, htmlDocumentUri);
+        const issueCreator = getReportIssueCreator(api);
+
+        // Act
+        const razorContent = await issueCreator.getProjectedHtml(razorDocument);
+
+        // Assert
+        assert.ok(razorContent.indexOf(expectedVSCodeHtmlContent) > 0);
+        assert.ok(razorContent.indexOf(expectedProjectedHtmlContent) > 0);
+    });
+
+    it('getProjectedHtml returns only projected Html if cannot locate vscodes text document', async () => {
+        // Arrange
+        const api = createTestVSCodeApi();
+        const htmlDocumentUri = api.Uri.parse('C:/path/to/file.cshtml.__virtual.html');
+        const hostDocumentUri = api.Uri.parse('C:/path/to/file.cshtml');
+        const expectedProjectedHtmlContent = 'Projected Html content';
+        const razorDocument = new TestRazorDocument(hostDocumentUri, hostDocumentUri.path);
+        razorDocument.htmlDocument = new TestProjectedDocument(expectedProjectedHtmlContent, htmlDocumentUri);
+        const issueCreator = getReportIssueCreator(api);
+
+        // Act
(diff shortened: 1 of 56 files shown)


src/Microsoft.AspNetCore.Razor.VSCode/package.json
```diff
@@ -4,13 +4,15 @@
...
+    "@types/clipboardy": "1.1.0",
     "tslint": "^5.11.0",
     "typescript": "^3.0.3",
     "vscode": "^1.1.22"
   },
   "dependencies": {
     "vscode-html-languageservice": "^2.1.7",
-    "vscode-languageclient": "^5.1.0"
+    "vscode-languageclient": "^5.1.0",
+    "clipboardy": "1.2.3"
```
Eilon: OSS good to go.

rchande:
@NTaylorMullen Nice, this is a pretty cool idea. We also ran into the GitHub limitation not having a `new issue` API allowing arbitrary length text. I don't remember if we added the ability to automatically copy our generated issues though (@akshita31)? 

It's interesting to me that you need specific gestures to start and stop logging. For us, most issues have a simple repro or are discovered during O# startup (missing SDKs, etc.). 

When this is running in the C# extension, how will we help users understand when they need to use `C#: report an issue` versus `Razor: report an issue`?

akshita31:
For the C# extension we invoke vscode.open on the github new issues url with the body that contains the information from the users system. However it has the restriction that we cannot put the logs etc into it that is why we have to ask users to paste the omnisharp log explicitly.

For the clipboard thing, I remember a discussion to add the capability for the extensions to access the clipboard but i am not sure if it has been added yet : https://github.com/Microsoft/vscode/issues/217

akshita31:
Also the "Start" and "Stop" for the logging collection seems a really cool idea 💯

NTaylorMullen:
> It's interesting (but makes perfect sense) to me that you need specific gestures to start and stop logging. For us, most issues have a simple repro or are discovered during O# startup (missing SDKs, etc.).

I wanted to reboot the Razor LanguageServer and start collecting immediately on reboot but I've yet to make a Razor Language Server reboot feasible (it doesn't work properly). If you guys were to have your "start collecting logs" reboot the server and capture those logs it could capture a lot of your concerns. Also, in our case we capture a lot of extra data a user wouldn't typically be able to easily capture such as the C# that VSCode sees vs the C# we see in our data type.

> When this is running in the C# extension, how will we help users understand when they need to use `C#: report an issue` versus `Razor: report an issue`?

Great question, I need to update our Razor log to provide instructions on how to start the file an issue wizard 😄. Today it just says to turn logs to verbose and file an issue on GitHub; can do better now! Also, it might make sense for me to forcibly launch the issue wizard with an error prompt saying "something unexpected happened, could you please repeat what you just did after pressing start?" or something like that when an unexpected error occurs.



> For the clipboard thing, I remember a discussion to add the capability for the extensions to access the clipboard but i am not sure if it has been added yet : [Microsoft/vscode#217](https://github.com/Microsoft/vscode/issues/217)

From my research (prior to using clipbordy) I found that it was added as a proposed API and requires flags from users in order to use it. Therefore, until it's available I decided to use cipboardy since it should be relatively simple to swap out once an official API is available.

ajaybhargavb:
Nice idea! I suggest generating collapsible content for pasting in the issue.

<details><summary>CLICK ME</summary>
<p>
```c#
// yay, less noisy issues!
```
</p>
</details>

akshita31:
Also, if you care about mono version(i dont know if razor language server depends on it) you can consider adding that. In omnisharp we have seen some problems specific to a particular mono version.

NTaylorMullen:
Note to self: Remember to try this out with light theme. I think my white text will barf

NTaylorMullen:
🆙 📅 - No longer in design. Added a ton of tests, changed the test output a bit (attached the new full log output) and verified things don't get unreadable in light mode.

New Issue Content: 
[issueContent.txt](https://github.com/aspnet/Razor.VSCode/files/2701236/issueContent.txt)


Light Mode:
image

NTaylorMullen:
Not quite sure whta's happenign with the CI but it says the build can't be found O_o

natemcmaster:
Something seems to be busted in the github connection. 🤷‍♂️ I'll try re-authorizing the connection.

natemcmaster:
It looks like it was using the old Oauth credentials. I've re-authorized using the GitHub app integration instead.

NTaylorMullen:
> It looks like it was using the old Oauth credentials. I've re-authorized using the GitHub app integration instead.

Thanks Nate!

NTaylorMullen:
Fixed the build in the PR

client/unittest/ReportIssueCreator.test.ts
```diff
@@ -0,0 +1,291 @@
...
+        // Arrange
+        const api = createTestVSCodeApi();
+        const issueCreator = getReportIssueCreator(api);
+
+        // Act & Assert
+        assert.doesNotThrow(() => issueCreator.generateExtensionTable());
+    });
+});
+
+class TestReportIssueCreator extends ReportIssueCreator {
```
SteveSandersonMS: I find this pattern a bit odd. We're unit testing methods that are protected only so they can be unit tested. I know we have a habit of doing this in shipping packages where the API surface needs to be locked down, but it's possibly different here since nobody except for us sees these APIs at all.

Other ways we could structure this:

* Only unit test the naturally public API surface. If a behavior isn't observable via public APIs, then there's no reason to specify that behavior.
* Or, maybe make the methods `public`. Nothing apart from our code can even see it either way. Then we don't need the `TestReportIssueCreator` subclass.

I'm not suggesting this is a big problem! If you prefer to leave it as-is, I don't object. I'm just wondering if this is a case where some patterns are leaking out from C#/framework coding practices into a scenario where they aren't really applicable.

client/unittest/ReportIssueDataCollector.test.ts
```diff
@@ -0,0 +1,94 @@
...
+        const eventEmitterFactory = new TestEventEmitterFactory();
+        const logger = new RazorLogger(api, eventEmitterFactory, Trace.Off);
+        const eventEmitter = eventEmitterFactory.create<vscode.TextDocument>();
+        const dataCollector = new ReportIssueDataCollector(eventEmitter.event, logger);
+
+        // Act
+        dataCollector.start();
+
+        // Assert
+        const lastLog = razorOutputChannel[razorOutputChannel.length - 1].trim();
```
SteveSandersonMS: Super-minor nitpick: the `.trim()` isn't needed.

Also it's a little odd that the condition on the next line is `> 0` not `>= 0`. Are we deliberately specifying it must not be at the start of the string, or is the intention to check the substring is present anywhere?

I know this is not a big deal.

client/unittest/ReportIssueDataCollector.test.ts
```diff
@@ -0,0 +1,94 @@
...
+
+        // Act
+        dataCollector.stop();
+
+        // Assert
+        const lastLog = razorOutputChannel[razorOutputChannel.length - 1].trim();
+        assert.ok(lastLog.indexOf('Stopping') > 0);
+    });
+
+    it('start->stop->collect captures the last focused Razor document', async () => {
```
SteveSandersonMS: It's great to have tests for this kind of thing! Nice one.

src/Microsoft.AspNetCore.Razor.VSCode/src/Diagnostics/ReportIssueCreator.ts
```diff
@@ -0,0 +1,244 @@
...
+    // Protected for testing
+    protected getInstalledExtensions() {
+        const extensions: Array<vscode.Extension<any>> = this.vscodeApi.extensions.all
+            .filter(extension => extension.packageJSON.isBuiltin === false);
+
+        return extensions.sort(this.sortExtensions);
+    }
+
+    // Protected for testing
+    protected sortExtensions(a: vscode.Extension<any>, b: vscode.Extension<any>): number {
```
SteveSandersonMS: TBH this feels a bit overcomplicated. Instead of having the `sortExtensions` method and all the unit tests for it, consider replacing L193 with:

```typescript
return extensions.sort((a, b) => 
    a.packageJSON.name.toLowerCase().localeCompare(
        b.packageJSON.name.toLowerCase()
    )
);
```

src/Microsoft.AspNetCore.Razor.VSCode/src/Diagnostics/ReportIssueCreator.ts
```diff
@@ -0,0 +1,244 @@
...
+    // Protected for testing
+    protected generateExtensionTable() {
+        const extensions = this.getInstalledExtensions();
+        if (extensions.length <= 0) {
+            return 'none';
+        }
+
+        const tableHeader = `|Extension|Author|Version|\n|---|---|---|`;
+        const table = extensions.map(
+            (e) => `|${e.packageJSON.name}|${e.packageJSON.publisher}|${e.packageJSON.version}|`).join(os.EOL);
```
SteveSandersonMS: We're using `\n` elsewhere, so it's odd to use `os.EOL` here.

src/Microsoft.AspNetCore.Razor.VSCode/src/Diagnostics/ReportIssueCreator.ts
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    private getDotnetInfo(): Promise<string> {
+        return new Promise<string>((resolve) => {
+            try {
+                cp.exec('dotnet --info', { cwd: process.cwd(), maxBuffer: 500 * 1024 }, (error, stdout, stderr) => {
+                    if (error) {
+                        throw error;
+                    } else if (stderr && stderr.length > 0) {
+                        throw new Error(stderr);
```
SteveSandersonMS: I don't think we should `throw` either here or on L232. Nothing will catch the async exception. I think we should use `reject` instead (also changing L228 to `return new Promise<string>((resolve, reject) => {`).

src/Microsoft.AspNetCore.Razor.VSCode/src/Diagnostics/ReportIssuePanel.ts
```diff
@@ -0,0 +1,170 @@
...
+                        this.issueContent = await this.reportIssueCreator.create(collectionResult);
+                        this.dataCollector = undefined;
+                    }
+
+                    await clipboardy.write(this.issueContent);
+                    vscode.window.showInformationMessage('Razor issue copied to clipboard');
+                    return;
+                case 'startIssue':
+                    this.issueContent = undefined;
+                    this.dataCollector = this.dataCollectorFactory.create();
```
SteveSandersonMS: Should we also check if there was a previous `dataCollector` and if so, dispose it?

Otherwise, if the user clicks "start" repeatedly, they will be leaking `focusRegistration` and `logRegistration`.

As a side note, it's awkward that `start` is separate from the `ReportIssueDataCollector` constructor. It implies you could `start` multiple times, but if you do, that will leak as well. Would it be safer to make `start` implicit in the constructor?

client/unittest/ReportIssueCreator.test.ts
NTaylorMullen: > * Only unit test the naturally public API surface. If a behavior isn't observable via public APIs, then there's no reason to specify that behavior.

I try to avoid this approach because you end up writing super complex integration level tests that have to fill out every part of the API that is used. In particular, the VSCode case is extra concerning because a lot of the APIs can't be mocked or faked properly.



> * Or, maybe make the methods `public`. Nothing apart from our code can even see it either way. Then we don't need the `TestReportIssueCreator` subclass.

I made them protected instead of public to avoid production code misuses. Aka, if a new dev is typing dot and then getting a completion list of available methods I don't want them to be using methods that are only public for the sake of testing.

client/unittest/ReportIssueDataCollector.test.ts
NTaylorMullen: > Super-minor nitpick: the `.trim()` isn't needed.

👍 

> Also it's a little odd that the condition on the next line is `> 0` not `>= 0`. Are we deliberately specifying it must not be at the start of the string, or is the intention to check the substring is present anywhere?

I'm assuming it's not at the start of the line because we also expect that logs have a date time stamp prefix. That being said, these tests shouldn't be concerned with that so I'll definitely change it for the sake of clarity.

src/Microsoft.AspNetCore.Razor.VSCode/src/Diagnostics/ReportIssueCreator.ts
NTaylorMullen: Good catch. A lot of this code was pulled from OmniSharp and I replaced small pieces of it. This was an oversight which was half-replaced.

src/Microsoft.AspNetCore.Razor.VSCode/src/Diagnostics/ReportIssueCreator.ts
NTaylorMullen: Interesting, I figured invoking cp.exec in the promise would throw from the inside but I see what you're saying. Good call. I'll fix.

src/Microsoft.AspNetCore.Razor.VSCode/src/Diagnostics/ReportIssuePanel.ts
NTaylorMullen: > As a side note, it's awkward that `start` is separate from the `ReportIssueDataCollector` constructor. It implies you could `start` multiple times, but if you do, that will leak as well. Would it be safer to make `start` implicit in the constructor?

hmm, true I could definitely do that. When I wrote the first iteration of this the flow was slightly different which warranted a separate start. Now that's not the case i'll fix.

> Should we also check if there was a previous `dataCollector` and if so, dispose it?
> 
> Otherwise, if the user clicks "start" repeatedly, they will be leaking `focusRegistration` and `logRegistration`.

Great catch, yes we should!

src/Microsoft.AspNetCore.Razor.VSCode/src/Diagnostics/ReportIssueCreator.ts
```diff
@@ -0,0 +1,244 @@
...
+                    } else {
+                        resolve(stdout);
+                    }
+                });
+            } catch (error) {
+                resolve(`A valid dotnet installation could not be found: ${error}`);
```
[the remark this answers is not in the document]
SteveSandersonMS: Oh actually that's debatable. Maybe you don't want this failure to cause the upstream flow to fail too.

For optimal clarity, I'd recommend using `reject` here, but then also having a `try/catch` around the call to it so the upstream code can explicitly show that it's not going to abort its flow even if `getDotnetInfo` fails.

PR #261, opened 2018-12-20
Add unit test infrastructure for VSCode project.

NTaylorMullen:
- VSCode's API package only contains definitions for the various VSCode APIs. This PR implements a lot of those definitions in "not implemented" ways. Basically, we provide an implementation for an API that by default throws; the test then needs to override the expectation of the VSCode API.
- Added tests for `RazorLogger` to showcase what this looks like.
- Added a `Launch Unit Tests` profile to allow us to run unit tests from VSCode via F5. Also added the pre-requisite tasks to ensure running tests in VSCode forces a compile of both the client (the test code) and the src assets.
- Changed `RazorLogger` and a few dependent classes to use APIs that have implementations so we can properly test them.
- Removed old arithmetic unit test because we actually have a unit test that validates product code.
- Added unit test running to the default ./build.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -33,6 +33,23 @@
                 "${workspaceFolder}/client/out/test/**/*.js"
             ],
             "preLaunchTask": "client-npm-watch"
-        }
+        },
+        {
+            "type": "node",
+            "request": "launch",
+            "name": "Launch Unit Tests",
+            "program": "${workspaceFolder}/client/node_modules/mocha/bin/_mocha",
+            "args": [
+                "-u",
+                "tdd",
+                "--timeout",
+                "999999",
+                "--colors",
+                "${workspaceFolder}/client/out/unittest"
+            ],
+            "sourceMaps": true,
+            "internalConsoleOptions": "openOnSessionStart",
+            "preLaunchTask": "client-npm-compile"
+        },
     ]
 }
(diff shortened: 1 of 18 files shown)


.vscode/launch.json
```diff
@@ -33,6 +33,23 @@
                 "${workspaceFolder}/client/out/test/**/*.js"
             ],
             "preLaunchTask": "client-npm-watch"
-        }
+        },
+        {
```
NTaylorMullen: This approach also allows VSCode to auto-attach and debug the tests as well 😄

src/Microsoft.AspNetCore.Razor.VSCode/src/Trace.ts
```diff
@@ -3,12 +3,8 @@
  * Licensed under the MIT License. See License.txt in the project root for license information.
  * ------------------------------------------------------------------------------------------ */
 
-import * as assert from 'assert';
```
NTaylorMullen: Messed up diff. Just added a new Trace type that looks like the one from jsonrpc. We need this Trace type in order to write tests. Basically a lot of the VSCode APIs that are shipped don't have implementations outside of VSCode. Therefore, we need to build adapters that look like those types so we can properly test them.

test/ClientTests/ClientTests.csproj
```diff
@@ -7,11 +7,18 @@
             some other VSCode instance at the same time, which is fine in CI but inconvenient locally.
         -->
         <CallTarget Targets="RunFunctionalTests" Condition="'$(IncludeFunctionalTests)'=='true'" />
+        <CallTarget Targets="RunUnitTests" />
```
NTaylorMullen: This enables unit tests to run as part of ./build.cmd without a flag 😄

PR #262, opened 2018-12-21
Expand unit test infrastructure.

NTaylorMullen:
- This expansion mostly mocks out a few core types that are needed to do anythign with the VSCode APIs and also changes the programming model to manipulate the VSCode workspace, extensions and output channels.
- Also cleaned up a few missed changes in some src files.
- This PR also serves as a pre-requisite for the ReportIssuePanel PR.


.vscode/tasks.json
```diff
@@ -6,8 +6,8 @@
         {
             "label": "client-npm-compile",
             "dependsOn": [
-                "client-npm-compile:shell",
-                "client-npm-compile:package"
+                "client-npm-compile:package",
```
NTaylorMullen: Changed the order because if you changed APIs on the src side and then reacted on the test side the test project would continuously fail to build because it wouldn't see the new changes.

client/tsconfig.json
```diff
@@ -2,7 +2,7 @@
 	"compilerOptions": {
 		"baseUrl": ".",
 		"module": "commonjs",
-		"target": "es5",
+		"target": "es6",
```
NTaylorMullen: This was an odd one. We were targeting es6 in our src project but not our client project. This led to issues when having types inherit from each other across projects.

NTaylorMullen:
Looks like I missed a few porting pieces here based on the build failure. I'll take a look

NTaylorMullen:
Investigated why. Ya there needs to be more bits here rather than in the other PR. gonna leave this PR failing until I merge the other one into it 😄

Issue #209, opened 2018-10-19
Consume Razor ProjectSystem with import understanding


NTaylorMullen:
There are a lot of import related project system changes in flight that we should consume to simplify our code. It'll enable us to not make `_ViewImports.cshtml` "fake work". It'll also enable us to not have as many ties to `ProjectSnapshotManager.GetLoadedProject`

NTaylorMullen:
Marking as blocked because we need to wait for a new SDK to flow through the system before we can properly consume new Razor bits

PR #263, opened 2018-12-27
Update to latest Razor

NTaylorMullen:
- Latest Razor included a ton of changes regarding the project system. One of those was the concept of "related documents". These changes consume those and remove the workaround logic we had to respond to _ViewImport documents being parsed and then causing re-parses of impacted documents.
- Updated tests and test infrastructure to work with new Razor bits. The biggest change here is the new Razor bits depend on certain workspace services to be available.
- Fixed an issue I encountered with adding/deleting _ViewImports files. There is a slight race where if the server asks for a file right as its being deleted the client throws. Changed this behavior to return an empty document instead of propogating the exceptino to the server.
- Fixed an issue where the BackgroundDocumentGenerator would null ref. Couldn't figure out why so added some logging and added pre-cautionary measures which ensure the generator continues to parse even after an explosion scenario.

#209

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.LanguageServer/BackgroundDocumentGenerator.cs
+++ b/src/Microsoft.AspNetCore.Razor.LanguageServer/BackgroundDocumentGenerator.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
@@ -179,9 +180,6 @@ namespace Microsoft.AspNetCore.Razor.LanguageServer
             {
                 _foregroundDispatcher.AssertBackgroundThread();
 
-                // Timer is stopped.
-                _timer.Change(Timeout.Infinite, Timeout.Infinite);
-
                 OnStartingBackgroundWork();
 
                 KeyValuePair<string, DocumentSnapshot>[] work;
@@ -233,7 +231,11 @@ namespace Microsoft.AspNetCore.Razor.LanguageServer
             catch (Exception ex)
             {
                 // This is something totally unexpected, let's just send it over to the workspace.
+                _logger.LogError("Unexpected error processing document: " + ex.Message);
                 ReportError(ex);
+
+                _timer?.Dispose();
+                _timer = null;
             }
         }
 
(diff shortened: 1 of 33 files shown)


src/Microsoft.AspNetCore.Razor.LanguageServer/Microsoft.AspNetCore.Razor.LanguageServer.csproj
```diff
@@ -14,32 +14,39 @@
...
   </ItemGroup>
 
   <Target Name="PublishLanguageServerNativeExecutables" AfterTargets="Pack" DependsOnTargets="PublishAllRids" />
 
   <Target Name="IncludeOmniSharpPlugin" AfterTargets="Publish" Condition="Exists('$(PublishDir)')">
-    <MSBuild
-      Projects="..\Microsoft.AspNetCore.Razor.OmniSharpPlugin\Microsoft.AspNetCore.Razor.OmniSharpPlugin.csproj"
-      Properties="PublishDir=$(MSBuildProjectDirectory)\..\Microsoft.AspNetCore.Razor.OmniSharpPlugin\bin\$(Configuration)\net461\publish;RuntimeIdentifier="
-      Targets="Publish" />
+    <MSBuild Projects="..\Microsoft.AspNetCore.Razor.OmniSharpPlugin\Microsoft.AspNetCore.Razor.OmniSharpPlugin.csproj" Properties="PublishDir=$(MSBuildProjectDirectory)\..\Microsoft.AspNetCore.Razor.OmniSharpPlugin\bin\$(Configuration)\net461\publish;RuntimeIdentifier=" Targets="Publish" />
```
ajaybhargavb: Revert. Looked better before.

src/Microsoft.AspNetCore.Razor.LanguageServer/BackgroundDocumentGenerator.cs
```diff
@@ -179,9 +180,6 @@ private async void Timer_Tick(object state)
             {
                 _foregroundDispatcher.AssertBackgroundThread();
 
-                // Timer is stopped.
-                _timer.Change(Timeout.Infinite, Timeout.Infinite);
```
ajaybhargavb: If this null-refing line is entirely gone, what is the point of the extra logging in the catch? Or was this a necessary change?

PR #277, opened 2019-01-11
Publish Razor parse diagnostics.

NTaylorMullen:
- We now listen for when a document has been processed by the `BackgroundDocumentGenerator` and on a cadence refresh diagnostics when they change.
- The `LanguageClient` already expects diagnostics to be flowed to the client and handles the publish diagnostics notifications appropriately.
- Because we can't proactively tell when a document has been closed we poll the snapshot manager on a 5 second interval re-checking all of our active documents.
- Add a RazorDiagnostic converter because we need to go from `RazorDiagnostic` to an LSP diagnostic. However, we only ever convert to an LSP diagnostic when we've found new diagnostics to publish.
- Added tests to validate the `RazorDiagnosticConverter` and `RazorDiagnosticPublisher` work as expected.

image

#30

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.LanguageServer/RazorDiagnosticsPublisher.cs
+++ b/src/Microsoft.AspNetCore.Razor.LanguageServer/RazorDiagnosticsPublisher.cs
@@ -1,0 +1,234 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.CodeAnalysis.Razor;
+using Microsoft.CodeAnalysis.Razor.ProjectSystem;
+using Microsoft.Extensions.Logging;
+using OmniSharp.Extensions.LanguageServer.Protocol.Models;
+using OmniSharp.Extensions.LanguageServer.Protocol.Server;
+
+namespace Microsoft.AspNetCore.Razor.LanguageServer
+{
+    internal class RazorDiagnosticsPublisher : DocumentProcessedListener
+    {
+        // Internal for testing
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNetCore.Razor.LanguageServer/RazorDiagnosticsPublisher.cs
```diff
@@ -0,0 +1,235 @@
...
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.CodeAnalysis.Razor;
+using Microsoft.CodeAnalysis.Razor.ProjectSystem;
+using Microsoft.Extensions.Logging;
+using OmniSharp.Extensions.LanguageServer.Protocol.Models;
+using OmniSharp.Extensions.LanguageServer.Protocol.Server;
+
+namespace Microsoft.AspNetCore.Razor.LanguageServer
+{
+    internal class RazorDiagnosticsPublisher : DocumentProcessedListener
```
ajaybhargavb: I feel like this is too specific. I'd imagine we might do other things in the future that listen to document processed events. I suggest having a `DefaultDocumentProcessedListener` and having that call into `RazorDiagnosticPublisher`.

NTaylorMullen: Not quite sure I follow the suggestion. Could you clarify a bit more?

ajaybhargavb: Imagine we want to implement another feature for which we need to know when the document processing is complete. How would you hook that up? Since `DocumentProcessedListener` is a singleton, we would already have an implementation of it (`RazorDiagnosticPublisher`) registered.

NTaylorMullen: Ahh, you missed this piece: https://github.com/aspnet/Razor.VSCode/pull/277/files#diff-71b7bdf65ccf5bc35df26932bdc6b4dfR31

You can have multiple `DocumentProcessedListener`'s, it imports an `IEnumerable` of them 😄

ajaybhargavb: Ah true. This is fine then 😄

PR #289, opened 2019-02-13
Fix Html completion timeout on freshly opened documents.

NTaylorMullen:
- Prior to this we wouldn't be updating the html buffer with the documents content when the doc was opened. This means that if a completion request via ctrl + j was invoked before typing it would never work.
- Added a functional test to validate behavior.

### DIFF
--- a/client/test/Completions.test.ts
+++ b/client/test/Completions.test.ts
@@ -38,6 +38,25 @@ describe('Completions', () => {
         await pollUntil(() => vscode.window.visibleTextEditors.length === 0, 1000);
     });
 
+    it('Can get HTML completions on document open', async () => {
+        // This test relies on the Index.cshtml document containing at least 1 HTML tag in it.
+        // For the purposes of this test it locates that tag and tries to get the Html completion
+        // list from it.
+
+        const content = doc.getText();
+        const tagNameIndex = content.indexOf('<') + 1;
+        const docPosition = doc.positionAt(tagNameIndex);
+        const completions = await vscode.commands.executeCommand<vscode.CompletionList>(
+            'vscode.executeCompletionItemProvider',
+            doc.uri,
+            docPosition);
+        const matchingCompletions = completions!.items
+            .filter(item => (typeof item.insertText === 'string') && item.insertText === 'iframe')
+            .map(item => item.insertText as string);
+
+        assert.deepEqual(matchingCompletions, ['iframe']);
+    });
+
     it('Can complete C# code blocks', async () => {
         const lastLine = new vscode.Position(doc.lineCount - 1, 0);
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Razor.VSCode/src/RazorDocumentManager.ts
```diff
@@ -128,9 +128,10 @@ export class RazorDocumentManager implements IRazorDocumentManager {
         return document;
     }
 
-    private openDocument(uri: vscode.Uri) {
+    private async openDocument(uri: vscode.Uri) {
         const document = this._getDocument(uri);
 
+        this.updateHtmlBuffer(document);
```
NTaylorMullen: This is the primary fix, the other awaits were potential race conditions

client/test/Completions.test.ts
```diff
@@ -38,6 +38,25 @@ describe('Completions', () => {
...
+    it('Can get HTML completions on document open', async () => {
+        // This test relies on the Index.cshtml document containing at least 1 HTML tag in it.
+        // For the purposes of this test it locates that tag and tries to get the Html completion
+        // list from it.
+
+        const content = doc.getText();
+        const tagNameIndex = content.indexOf('<') + 1;
+        const docPosition = doc.positionAt(tagNameIndex);
+        const completions = await vscode.commands.executeCommand<vscode.CompletionList>(
+            'vscode.executeCompletionItemProvider',
```
ajaybhargavb: Since this is a functional test(from what you said in the description), wouldn't it be better to send a keystroke (ctrl+j) at this location instead of invoking the CompletionProvider ourselves?

src/Microsoft.AspNetCore.Razor.VSCode/src/RazorDocumentManager.ts
```diff
@@ -128,9 +128,10 @@ export class RazorDocumentManager implements IRazorDocumentManager {
         return document;
     }
 
-    private openDocument(uri: vscode.Uri) {
+    private async openDocument(uri: vscode.Uri) {
```
ajaybhargavb: Why async? There is nothing async going on here.

client/test/Completions.test.ts
NTaylorMullen: Wish we could but sadly can't T_T

src/Microsoft.AspNetCore.Razor.VSCode/src/RazorDocumentManager.ts
NTaylorMullen: Oh derp, forgot that I had removed the async pieces when investigating the fix 😲

NTaylorMullen:
🆙  📅

Issue #16, opened 2018-08-03
Initial TagHelper discovery.


NTaylorMullen:
Right now we have a `NoopTagHelperResolver` flowing through the Razor project engine; that prevents any discovery of TagHelpers. We'll need to develop a system for finding TagHelpers involved in the system on document open.

NTaylorMullen:
https://github.com/aspnet/Razor.VSCode/commit/72dc1564ed092d7c1d5761a3e3c11fd2d87d39c6

PR #290, opened 2019-02-14
Add support for TagHelper/Component parsing.

NTaylorMullen:
- Expanded the OmniSharpPlugin project to now spin up a project engine and use it to resolve TagHelpers in the context of the associated OmniSharp project.
- Updated the Razor dependency of all projects, this included consuming new changes for `ProjectWorkspaceState` which enables us to replace TagHelpers for a given project without needing a Roslyn project.
- Due to strong naming restrictions had to `Assembly.Load` and reflectively invoke several classes, this included the TagHelperResolver.
- Since we now have minimal project engines being spun up in the OmniSharp side of the world added a LanguageServer.Common project that encompasses the minimal logic needed to build a proper project engine.
- Expanded our language querying mechanism to understand TagHelpers better. If we find a non-TagHelper classified span we use it to determine the content type, otherwise we fallback to TagHelper classified spans to determine if an area should be treated as Html or not.
- Added tests where I could. Because of how the OmniSharp plugin works was not able to add a TagHelper functional test to validate the correctness of TagHelpers. This is because our functional tests use the currently released OmniSharp which does not ship the appropriate OmniSharp plugin that understands TagHelpers. We also can't force the OmniSharp server to load the new plugin because we can't accurately resolve the path at configuration time to tell OmniSharp early enough which path to use for anyone who clones the Razor.VSCode repo.

#16


src/Microsoft.AspNetCore.Razor.LanguageServer.Common/ProjectEngineFactory_1_1.cs
```diff
@@ -15,7 +15,7 @@ internal class ProjectEngineFactory_1_1 : IProjectEngineFactory
         public RazorProjectEngine Create(RazorConfiguration configuration, RazorProjectFileSystem fileSystem, Action<RazorProjectEngineBuilder> configure)
         {
             // Rewrite the assembly name into a full name just like this one, but with the name of the MVC design time assembly.
-            var assemblyName = new AssemblyName(typeof(ProjectEngineFactory_1_1).Assembly.FullName);
+            var assemblyName = new AssemblyName(typeof(RazorProjectEngine).Assembly.FullName);
```
NTaylorMullen: We do the `typeof(XYZ)` in order to derive the strong named key + version of the assembly to load. Since we're loading a Razor binary we should use its version/key instead of our own.

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/DefaultProjectEngineFactory.cs
```diff
@@ -0,0 +1,48 @@
...
+        public override RazorProjectEngine Create(
+            RazorConfiguration configuration,
+            RazorProjectFileSystem fileSystem,
+            Action<RazorProjectEngineBuilder> configure)
+        {
+            lock (InitializationLock)
+            {
+                if (_defaultEngineFactory == null)
+                {
+                    var razorWorkspaces = Assembly.Load("Microsoft.CodeAnalysis.Razor.Workspaces");
```
NTaylorMullen: In this assembly, all this reflection loading magic is needed due to strong naming restrictions 😢. The OmniSharpPlugin dll can't be strong named because OmniSharp isn't. Therefore, we can't IVT to it from Razor.

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/ProjectLoadListener.cs
```diff
@@ -78,12 +105,20 @@ public void ProjectLoaded(ProjectLoadedEventArgs args)
                     return;
                 }
 
-                var razorConfiguration = GetRazorConfiguration(args.ProjectInstance);
+                var razorConfiguration = GetRazorConfiguration(args.ProjectInstance, _projectConfigurationProviders);
+                var projectDirectory = args.ProjectInstance.GetPropertyValue(MSBuildProjectDirectoryPropertyName);
+                var fileSystem = RazorProjectFileSystem.Create(projectDirectory);
```
NTaylorMullen: Thought about adding tests for all these guys but it ends up just being a boatload of ceremony with not a lot of value.

NTaylorMullen:
As a future note for testing since the currently released version of OmniSharp can't load the plugin in this PR. You need to add the following settings to your dev project when verifying:
```JSON
{
    "razor.devmode": true,
    "razor.plugin.path": "C:/GitHub/Razor.VSCode/src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/bin/Debug/net461/Microsoft.AspNetCore.Razor.OmniSharpPlugin.dll",
    "omnisharp.path": "1.32.9-beta.56"
}
```

test/Microsoft.AspNetCore.Razor.LanguageServer.Test/RazorLanguageEndpointTest.cs
```diff
@@ -334,12 +392,14 @@ public void GetCSharpProjectedPosition_CSharp_OnTrailingEdge()
             Assert.Equal(30, projectedPositionIndex);
         }
 
-        private IReadOnlyList<ClassifiedSpanInternal> GetClassifiedSpans(string text)
+        private (IReadOnlyList<ClassifiedSpanInternal> classifiedSpans, IReadOnlyList<TagHelperSpanInternal> tagHelperSpans) GetClassifiedSpans(string text, IReadOnlyList<TagHelperDescriptor> tagHelpers = null)
         {
-            var codeDocument = CreateCodeDocument(text);
+            var codeDocument = CreateCodeDocument(text, tagHelpers);
+            codeDocument.SetTagHelpers(tagHelpers);
```
NTaylorMullen: no need to set taghelpers

NTaylorMullen:
Merging this to unblock myself, feel free to review afterwards and will address comments then.

src/Microsoft.AspNetCore.Razor.LanguageServer.Common/ProjectEngineFactory_3_0.cs
```diff
@@ -2,28 +2,35 @@
...
             var initializer = extension.CreateInitializer();
 
             return RazorProjectEngine.Create(configuration, fileSystem, b =>
             {
+                CompilerFeatures.Register(b);
+
                 initializer.Initialize(b);
                 configure?.Invoke(b);
+
+                b.Features.OfType<ComponentDocumentClassifierPass>().Single().MangleClassNames = true;
```
ajaybhargavb: This was the line that throws for this [bug](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/788091). Did you run into this at any point?

NTaylorMullen: This line has thrown a few times. That being said I've been repairing the airplane while it's flying so when I encountered the issue I was in the midst of fixing other issues and didn't directly correlate the exception (I assumed it was my other work). The exception stopped and I haven't ran into it for a while.

Issue #93, opened 2018-08-27
Simple TagHelper completion


NTaylorMullen:
This is a marker issue for a potential Razor.VSCode alpha implementation of TagHelpers. It wouldn't require a deep understanding of the Html representation of the document and would enable scenarios such as Blazor. 

Completion of this issue would entail providing contextless attribute and element completion. Meaning TagHelpers that have no parent or attribute restrictions.

https://github.com/aspnet/Razor.VSCode/issues/18 is the full TagHelper completion story once we have more information about the SyntaxTree.

NTaylorMullen:
@ajaybhargavb I'll start poking at this one now in parallel with the other TagHelper features

NTaylorMullen:
668a4c4d84dc8b05b3eb80d51763601a1f8e8ffa
1b92523407351943013040a039b35af9ae7ccc80

PR #293, opened 2019-02-21
Add TagHelper element completion.

NTaylorMullen:
- This does not include any description/documentation support (will be in a coming PR).
- Added exhaustive tests to ensure the completion works as expected.
- Integrated TagHelper completion into the language server's completion endpoint so when VSCode prompts for completions we do a fork to see if there are any TagHelper completions available for the location.

image

#93

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.LanguageServer/DefaultTagHelperCompletionService.cs
+++ b/src/Microsoft.AspNetCore.Razor.LanguageServer/DefaultTagHelperCompletionService.cs
@@ -306,6 +306,9 @@ namespace Microsoft.AspNetCore.Razor.LanguageServer
                     Kind = CompletionItemKind.TypeParameter,
                     CommitCharacters = ElementCommitCharacters,
                 };
+                var tagHelperDescriptions = completion.Value.Select(tagHelper => new TagHelperDescriptionInfo(tagHelper.GetTypeName(), tagHelper.Documentation));
+                var elementDescription = new ElementDescriptionInfo(tagHelperDescriptions.ToList());
+                razorCompletionItem.SetDescriptionData(elementDescription);
 
                 completionItems.Add(razorCompletionItem);
             }
--- a/src/Microsoft.AspNetCore.Razor.LanguageServer/DefaultTagHelperCompletionService.cs
+++ b/src/Microsoft.AspNetCore.Razor.LanguageServer/DefaultTagHelperCompletionService.cs
@@ -268,6 +268,13 @@ namespace Microsoft.AspNetCore.Razor.LanguageServer
                     Kind = CompletionItemKind.TypeParameter,
                     CommitCharacters = AttributeCommitCharacters,
                 };
+                var attributeDescriptions = completion.Value.Select(boundAttribute => new TagHelperAttributeDescriptionInfo(
+                    boundAttribute.DisplayName,
+                    boundAttribute.GetPropertyName(),
+                    indexerCompletion ? boundAttribute.IndexerTypeName : boundAttribute.TypeName,
+                    boundAttribute.Documentation));
+                var attributeDescriptionInfo = new AttributeDescriptionInfo(attributeDescriptions.ToList());
+                razorCompletionItem.SetDescriptionInfo(attributeDescriptionInfo);
 
                 completionItems.Add(razorCompletionItem);
             }
@@ -308,7 +315,7 @@ namespace Microsoft.AspNetCore.Razor.LanguageServer
                 };
                 var tagHelperDescriptions = completion.Value.Select(tagHelper => new TagHelperDescriptionInfo(tagHelper.GetTypeName(), tagHelper.Documentation));
                 var elementDescription = new ElementDescriptionInfo(tagHelperDescriptions.ToList());
-                razorCompletionItem.SetDescriptionData(elementDescription);
+                razorCompletionItem.SetDescriptionInfo(elementDescription);
 
                 completionItems.Add(razorCompletionItem);
             }
--- a/src/Microsoft.AspNetCore.Razor.LanguageServer/DefaultTagHelperCompletionService.cs
+++ b/src/Microsoft.AspNetCore.Razor.LanguageServer/DefaultTagHelperCompletionService.cs
@@ -16,6 +16,7 @@ namespace Microsoft.AspNetCore.Razor.LanguageServer
 {
     internal class DefaultTagHelperCompletionService : TagHelperCompletionService
     {
+        private static readonly Container<string> AttributeCommitCharacters = new Container<string>(" ");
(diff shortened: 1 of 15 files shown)


src/Microsoft.AspNetCore.Razor.LanguageServer/DefaultTagHelperCompletionService.cs
```diff
@@ -0,0 +1,299 @@
...
+        {
+            var stringifiedAttributes = new List<KeyValuePair<string, string>>();
+
+            for (var i = 0; i < attributes.Count; i++)
+            {
+                var attribute = attributes[i];
+                if (attribute is MarkupTagHelperAttributeSyntax tagHelperAttribute)
+                {
+                    var name = tagHelperAttribute.Name.GetContent();
+                    var value = tagHelperAttribute.Value.GetContent();
```
ajaybhargavb: Is Value guaranteed to not be null? (Probably more of a question to myself 😆). In any case, might be better to be defensive.

src/Microsoft.AspNetCore.Razor.LanguageServer/RazorCompletionEndpoint.cs
```diff
@@ -107,13 +115,19 @@ public async Task<CompletionList> Handle(CompletionParams request, CancellationT
                     Detail = razorCompletionItem.Description,
                     Documentation = razorCompletionItem.Description,
                     FilterText = razorCompletionItem.DisplayText,
-                    SortText = razorCompletionItem.DisplayText,
+                    SortText = "__Razor__" + razorCompletionItem.DisplayText,
```
ajaybhargavb: What is this for?

src/Microsoft.AspNetCore.Razor.LanguageServer/RazorCompletionEndpoint.cs
```diff
@@ -125,7 +139,7 @@ public CompletionRegistrationOptions GetRegistrationOptions()
             {
                 DocumentSelector = RazorDefaults.Selector,
                 ResolveProvider = true,
-                TriggerCharacters = new Container<string>("@"),
+                TriggerCharacters = new Container<string>("@", "<"),
```
ajaybhargavb: Ooh interesting. Will this be triggered for any `<` in the document even in CSharp context?

src/Microsoft.AspNetCore.Razor.LanguageServer/DefaultTagHelperCompletionService.cs
```diff
@@ -0,0 +1,299 @@
...
+using Microsoft.AspNetCore.Razor.Language.Syntax;
+using Microsoft.VisualStudio.Editor.Razor;
+using OmniSharp.Extensions.LanguageServer.Protocol.Models;
+using RazorTagHelperCompletionService = Microsoft.VisualStudio.Editor.Razor.TagHelperCompletionService;
+
+namespace Microsoft.AspNetCore.Razor.LanguageServer
+{
+    internal class DefaultTagHelperCompletionService : TagHelperCompletionService
+    {
+        private static readonly Container<string> ElementCommitCharacters = new Container<string>(" ");
```
ajaybhargavb: What is a commit character?

NTaylorMullen: When the completion is selected it's the character you can press to insert the completion. I.e. for C# typing `@Dat` and then pressing `.` results in `@DateTime.`

src/Microsoft.AspNetCore.Razor.LanguageServer/RazorCompletionEndpoint.cs
NTaylorMullen: When VSCode has all the completion items it controls what order they show up in alphabetically. Means you could have `@page` show up first even though it begins with p if it's sort text starts with an A. In this case I shouldn't be changing the sort text. Will undo

src/Microsoft.AspNetCore.Razor.LanguageServer/RazorCompletionEndpoint.cs
NTaylorMullen: Sadly yes. Therefore a C# completion list will come up; however, no TagHelper or Html completions will be available because it'll be an invalid Html/TagHelper context.

src/Microsoft.AspNetCore.Razor.LanguageServer/DefaultTagHelperCompletionService.cs
ajaybhargavb: I see. What about `Enter` and `Tab`?

src/Microsoft.AspNetCore.Razor.LanguageServer/RazorCompletionEndpoint.cs
ajaybhargavb: Is it something omnisharp has to support if we want to do it right?

src/Microsoft.AspNetCore.Razor.LanguageServer/DefaultTagHelperCompletionService.cs
NTaylorMullen: Those are defaults 😄

src/Microsoft.AspNetCore.Razor.LanguageServer/RazorCompletionEndpoint.cs
NTaylorMullen: Not entirely, we'll have to special case the scenario, I filed an issue: https://github.com/aspnet/Razor.VSCode/issues/295

src/Microsoft.AspNetCore.Razor.LanguageServer/DefaultTagHelperCompletionService.cs
```diff
@@ -0,0 +1,299 @@
...
+                    stringifiedAttributes.Add(new KeyValuePair<string, string>(name, string.Empty));
+                }
+            }
+
+            return stringifiedAttributes;
+        }
+
+        private static (string ancestorTagName, bool ancestorIsTagHelper) GetNearestAncestorTagInfo(IEnumerable<SyntaxNode> ancestors)
+        {
+            foreach (var ancestor in ancestors)
```
ajaybhargavb: Spoke offline. The ancestor of a start tag can't be another start tag (Except in malformed cases). Should check for `MarkupElement` or `MarkupTaghelperElement` and get its `StartTag` property.

src/Microsoft.AspNetCore.Razor.LanguageServer/DefaultTagHelperCompletionService.cs
NTaylorMullen: Should add `>` to the list

Issue #18, opened 2018-08-03
TagHelper Completion Documentation


NTaylorMullen:
Allow TagHelper completions to flow documentation

NTaylorMullen:
60a9f1c6df27bd5bbcda5f3674ed9635caa6dfd7
c9676843d554bd1e8136b029e334957070f341a3

PR #296, opened 2019-02-22
Add TagHelper element completion descriptions.

NTaylorMullen:
- Completion documentation is expected to be fully populated once a user selects a completion item in the drop down list. At this point VSCode performs a completion item resolve request to the language server in order to populate any documentation. This is essential because populating the documentation usually requires a significant amount of work and is not something we want users waiting for. Therefore, in these changes we store just enough information on the completion item to re-hydrate a markdown based documentation for when one of our TagHelper completions is resolved.
- Given VSCodes limited viewport bandwidth our description factory has to do a lot of work to slim down type names in crefs and TagHelper typenames. Otherwise you end up with a description that never fits the box.
- Added exhaustive tests to ensure the completion descriptions populate correctly. Did only a few integration level tests for the completion description to make things easier to maintain. Didn't want to add a ton of work every time we decided to change what the completion descriptions looked like.
- Updated the completion endpoint to also be a resolve handler. Resolve handlers are responsible for populating documentation when a user selects a completion item.
- Added completion item extensions. We need to interact with completion items in two different services in order to store/retrieve description data, wanted to unify the way that was done with the extensions.

image

#18

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.LanguageServer/DefaultTagHelperDescriptionFactory.cs
+++ b/src/Microsoft.AspNetCore.Razor.LanguageServer/DefaultTagHelperDescriptionFactory.cs
@@ -13,6 +13,24 @@ namespace Microsoft.AspNetCore.Razor.LanguageServer
     {
         private static readonly Lazy<Regex> ExtractCrefRegex = new Lazy<Regex>(
             () => new Regex("<(see|seealso)[\\s]+cref=\"([^\">]+)\"[^>]*>", RegexOptions.Compiled, TimeSpan.FromSeconds(1)));
+        private static readonly IReadOnlyDictionary<string, string> PrimitiveDisplayTypeNameLookups = new Dictionary<string, string>(StringComparer.Ordinal)
+        {
+            [typeof(byte).FullName] = "byte",
+            [typeof(sbyte).FullName] = "sbyte",
+            [typeof(int).FullName] = "int",
+            [typeof(uint).FullName] = "uint",
+            [typeof(short).FullName] = "short",
+            [typeof(ushort).FullName] = "ushort",
+            [typeof(long).FullName] = "long",
+            [typeof(ulong).FullName] = "ulong",
+            [typeof(float).FullName] = "float",
+            [typeof(double).FullName] = "double",
+            [typeof(char).FullName] = "char",
+            [typeof(bool).FullName] = "bool",
+            [typeof(object).FullName] = "object",
+            [typeof(string).FullName] = "string",
+            [typeof(decimal).FullName] = "decimal",
+        };
 
         public override bool TryCreateDescription(ElementDescriptionInfo elementDescriptionInfo, out string markdown)
         {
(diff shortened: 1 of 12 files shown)


src/Microsoft.AspNetCore.Razor.LanguageServer/DefaultTagHelperDescriptionFactory.cs
```diff
@@ -0,0 +1,259 @@
...
+        {
+            var associatedTagHelperInfos = elementDescriptionInfo.AssociatedTagHelperDescriptions;
+            if (associatedTagHelperInfos.Count == 0)
+            {
+                markdown = null;
+                return false;
+            }
+
+            var descriptionBuilder = new StringBuilder();
+            for (var i = 0; i < associatedTagHelperInfos.Count; i++)
```
ajaybhargavb: I would be nice to add a comment here with an example of what this will look like.

src/Microsoft.AspNetCore.Razor.LanguageServer/DefaultTagHelperDescriptionFactory.cs
```diff
@@ -0,0 +1,259 @@
...
+                    summaryBuilder.Insert(cref.Index, $"`{reducedValue}`");
+                }
+            }
+            var lines = summaryBuilder.ToString().Split(new[] { '\n' }, StringSplitOptions.None).Select(line => line.Trim());
+            var finalSummaryContent = string.Join(Environment.NewLine, lines);
+            return finalSummaryContent;
+        }
+
+        // Internal for testing
+        internal static bool TryExtractSummary(string documentation, out string summary)
```
ajaybhargavb: Curious. Would it have made your life easier if you used Roslyn to parse these documentation?

test/Microsoft.AspNetCore.Razor.LanguageServer.Test/DefaultTagHelperDescriptionFactoryTest.cs
```diff
@@ -0,0 +1,381 @@
...
+            var reduced = DefaultTagHelperDescriptionFactory.ReduceMemberName(content);
+
+            // Assert
+            Assert.Equal("SomeType<Foo.Bar<Baz,Fi>>.SomeMethod(Foo.Bar<System.String>,Baz<Something>.Fi)", reduced);
+        }
+
+        [Theory]
+        [InlineData("Microsoft.AspNetCore.SometTagHelpers.SomeType.Foo.Bar<Baz.Phi>>")]
+        [InlineData("Microsoft.AspNetCore.SometTagHelpers.SomeType.Foo.Bar{Baz.Phi}}")]
+        [InlineData("Microsoft.AspNetCore.SometTagHelpers.SomeType.Foo.Bar(Baz.Phi))")]
```
ajaybhargavb: Add more malformed cases here. Like `Foo..Bar`, `Foo{.>` etc. There are probably not realistic scenarios but doesn't hurt to make sure we don't explode.

src/Microsoft.AspNetCore.Razor.LanguageServer/DefaultTagHelperDescriptionFactory.cs
NTaylorMullen: Yup but they need a Roslyn solution in order to parse it correctly. Something we don't have on the LanguageServer side.

test/Microsoft.AspNetCore.Razor.LanguageServer.Test/DefaultTagHelperDescriptionFactoryTest.cs
NTaylorMullen: Can add your second scenario. The first one we'd actually return a reduced member because we don't care if the type name is string.Empty. (multiple dots are recoverable)

Issue #17, opened 2018-08-03
TagHelper refresh


NTaylorMullen:
On build, on a Blazor component etc. We need to understand when we need to refresh TagHelpers and forcefully update documents.

NTaylorMullen:
This is dependent on https://github.com/aspnet/AspNetCore/issues/6672

NTaylorMullen:
c45160be9ee1e3176cb4c85c9612fcaa090b6b82
304e6c43096e33af5e1f389b6c44ef1ef84320e5

PR #300, opened 2019-03-05
Add a Razor project system to OmniSharp plugin.

NTaylorMullen:
- The goal behind this change is to get the OmniSharp plugin working on a model of maintaining project state for all of the various Razor projects in the workspace. This means when a project change event happens we either add or update the corresponding ProjectSnapshot. At which point we have a separate service that listens for project changes and then kicks off TagHelper resolution in the background to fill in TagHelpers and other workspace state information. By using this model we'll be able to have 1 serialization model (`ProjectSnapshot`) that represents everything we need to know about the project; it also enables us to consume new Razor features in the future (i.e. root namespace for a project, specific Razor documents for a project etc.).
- Moved a lot more infrastructure and dependencies into `Microsoft.AspNetCore.Razor.LanguageServer.Common`. The goal was to re-use as much project system glue code as possible between the language server and the OmniSharp plugin.
- Added a `Microsoft.AspNetCore.Razor.Test.Common` project to share all test related infrastructure. Now that the plugin also uses the project system we need to be able to re-use all of the snapshot test code that we have.
- Added a `Microsoft.AspNetCore.Razor.OmniSharpPlugin.StrongNamed` project to act as a shim between strong  named and non-strong named bits. The OmniSharp plugin cannot be strong named signed which means we can't have IVT to Razor's project system types. To make use of the code that already exists in Razor I had to shim out the important project system bits as public and then re-consume them in the OmniSharp plugin.
- Added additional serialization logic to enable proper serialization of a project snapshot down to disk and then reconsumption in the language server.
- Removed TargetFramework understanding in the serialized project file. It was only ever used for telemetry and that knowledge is moving over into OmniSharp.
- Fixed a bug in the foreground dispatcher where awaited statements wouldn't return to the proper thread after dispatching to the foreground thread.
- Added a project change publisher to listen for project change events on the OmniSharp plugin side of the world. When a project change event occurs we take that opportunity to serialize the project down into the workspace so the language server can pick it up.
- Added tests to verify a lot of the project system functionality in the OmniSharp world. This was a bit ugly because in order to properly construct anything I needed to use reflection to instantiate my way to victory.
- Extended Razor's `WorkspaceProjectStateChangeDetector` to be more free-threaded when being invoked from the OmniSharp workspace. Ultimately this was to work around an issue in OmniSharp's language server where they mutate the C# workspace on any thread (unsupported).
- Added an assembly service aggregator to enable us to declare Roslyn workspace/language services in the strong named assembly.
- Re-exported MEF services from the OmniSharp strong named assembly in the OmniSharp plugin assembly. This was required because there's no hook to enable OmniSharp to also harvest MEF services from the strong named plugin assembly.

Pre-requisite to #17


src/Microsoft.AspNetCore.Razor.LanguageServer.Common/AdhocLanguageServices.cs
```diff
@@ -0,0 +1,49 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.CodeAnalysis.Host;
+using Microsoft.CodeAnalysis.Razor;
+
+namespace Microsoft.AspNetCore.Razor.LanguageServer.Common
+{
+    public class AdhocLanguageServices : HostLanguageServices
```
NTaylorMullen: Pulled out all of these adhoc services from the LanguageServer's project snapshot manager accessor. Direct Cut & Paste

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin.StrongNamed/OmniSharpWorkspaceProjectStateChangeDetector.cs
```diff
@@ -0,0 +1,116 @@
...
+            {
+                throw new ArgumentNullException(nameof(foregroundDispatcher));
+            }
+
+            if (workspaceStateGenerator == null)
+            {
+                throw new ArgumentNullException(nameof(workspaceStateGenerator));
+            }
+
+            InternalWorkspaceProjectStateChangeDetector = new ForegroundWorkspaceProjectStateChangeDetector(
```
NTaylorMullen: This is what does TagHelper resolution now. Trying to re-use what core-Razor uses so when we change one the other one gets updated as well. Basically this guy listens for `Workspace` & `ProjectSnapshotManager` change events and then decides if it wants to update TagHelpers. 

[Here's the Razor bits](https://github.com/aspnet/AspNetCore-Tooling/blob/8ee676d84cea1a375f6179aaf2db86f87f73625d/src/Razor/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/WorkspaceProjectStateChangeDetector.cs).

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/DefaultProjectChangePublisher.cs
```diff
@@ -0,0 +1,131 @@
...
+using Microsoft.AspNetCore.Razor.OmniSharpPlugin.StrongNamed.Serialization;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+
+namespace Microsoft.AspNetCore.Razor.OmniSharpPlugin
+{
+    [Shared]
+    [Export(typeof(ProjectChangePublisher))]
+    [Export(typeof(OmniSharpProjectSnapshotManagerChangeTrigger))]
+    internal class DefaultProjectChangePublisher : ProjectChangePublisher, OmniSharpProjectSnapshotManagerChangeTrigger
```
NTaylorMullen: This is the piece that ties everything together in the plugin. It listens to change events from the snapshot manager and then serializes the projects down to the file system. Nice thing about this is that the project snapshot manager handles ensuring everything is synchronized correctly, doesn't re-dispatch on no-op changes, and overall just manages every bit of state we could ever want.

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/FilePathComparer.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.AspNetCore.Razor.OmniSharpPlugin
+{
+    internal static class FilePathComparer
```
NTaylorMullen: A direct copy from Razor.

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/MSBuildProjectLoadListener.cs
```diff
@@ -98,42 +107,24 @@ public async void ProjectLoaded(ProjectLoadedEventArgs args)
...
-                }
-                catch (Exception)
-                {
-                    // TODO: Add retry.
-                }
+                    var projectSnapshot = _projectManager.GetLoadedProject(projectFilePath);
+                    var hostProject = new OmniSharpHostProject(projectFilePath, razorConfiguration);
+                    if (projectSnapshot == null)
+                    {
+                        _projectManager.ProjectAdded(hostProject);
```
NTaylorMullen: Now all this guy does (instead of serializing to disk) is update the project system based on MSBuild events that flow.

test/Microsoft.AspNetCore.Razor.OmniSharpPlugin.Test/OmniSharpTestBase.cs
```diff
@@ -0,0 +1,54 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+using System.Reflection;
+using Microsoft.AspNetCore.Razor.Test.Common;
+
+namespace Microsoft.AspNetCore.Razor.OmniSharpPlugin
+{
+    public class OmniSharpTestBase : TestBase
```
NTaylorMullen: Tried to group every bit of ugliness we have to endure from interoping between strong-named/non-strong named assemblies here.

Embrace the hack!

src/Microsoft.AspNetCore.Razor.VSCode/src/TelemetryReporter.ts
```diff
@@ -74,7 +73,6 @@ export class TelemetryReporter {
             {
                 path: project.path,
                 configurationName,
-                targetFramework,
```
NTaylorMullen: FYI @akshita31 we'll need to coordinate your telemetry work with our next release (mid April maybe?). I'm removing the `TargetFramework` reporting from our side.

akshita31: @NTaylorMullen I am trying to fit that into our next release(sometime around next week), so we should be good

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/OmniSharpStrongNamedExports.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Composition;
+using Microsoft.AspNetCore.Razor.OmniSharpPlugin.StrongNamed;
+using OmniSharp;
+
+namespace Microsoft.AspNetCore.Razor.OmniSharpPlugin
+{
+    // We need to re-export MEF based services from the OmniSharp plugin strong named assembly in order
+    // to make those services available via MEF. This isn't an issue for Roslyn based services because
+    // we're able to hook into OmniSharp's Roslyn service aggregator to allow it to inspect the strong
+    // named plugin assembly.
```
ajaybhargavb: How long term is all this going to be? Is there a plan to remove all this hackery in the future?

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/ProjectSnapshotManagerInstantiator.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Composition;
+using OmniSharp.MSBuild.Notification;
+
+namespace Microsoft.AspNetCore.Razor.OmniSharpPlugin
+{
+    [Shared]
+    [Export(typeof(IMSBuildEventSink))]
+    internal class ProjectSnapshotManagerInstantiator : IMSBuildEventSink
+    {
+        // The entire purpose of this class is to ensure the project manager is instantiated.
```
ajaybhargavb: Could you expand this comment to include what layer will this be called and why we need it to be initialized at this layer?

test/Microsoft.AspNetCore.Razor.OmniSharpPlugin.Test/DefaultProjectChangePublisherTest.cs
```diff
@@ -0,0 +1,165 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.Extensions.Logging;
+using Xunit;
+using Xunit.Sdk;
+
+namespace Microsoft.AspNetCore.Razor.OmniSharpPlugin
```
ajaybhargavb: Lol what happened to Github coloring?

test/Microsoft.AspNetCore.Razor.OmniSharpPlugin.Test/DefaultProjectChangePublisherTest.cs
```diff
@@ -0,0 +1,165 @@
...
+            publisher.SetPublishFilePath(hostProject.FilePath, expectedPublishFilePath);
+            snapshotManager.ProjectAdded(hostProject);
+            var newConfiguration = RazorConfiguration.Create(RazorLanguageVersion.Experimental, "Custom", Enumerable.Empty<RazorExtension>());
+            var newHostProject = new OmniSharpHostProject(hostProject.FilePath, newConfiguration);
+
+            // Act
+            snapshotManager.ProjectConfigurationChanged(newHostProject);
+
+            // Assert
+            Assert.Equal(2, attemptsToSerialize);
```
ajaybhargavb: I don't understand what's going on here. Why is this attempted twice?

src/Microsoft.AspNetCore.Razor.LanguageServer.Common/VSCodeForegroundDispatcher.cs
```diff
@@ -40,7 +40,17 @@ private ForegroundTaskScheduler()
 
             protected override void QueueTask(Task task) => _tasks.Add(task);
 
-            protected override bool TryExecuteTaskInline(Task task, bool taskWasPreviouslyQueued) => TryExecuteTask(task);
+            protected override bool TryExecuteTaskInline(Task task, bool taskWasPreviouslyQueued)
+            {
+                // If the task was previously queued it means that we're ensuring it's running on our single threaded scheduler.
+                // Otherwise, we can't enforce that behavior and therefore need it to be re-queued before execution.
+                if (taskWasPreviouslyQueued)
```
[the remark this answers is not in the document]
NTaylorMullen:
```C#
Task.Factory.StartNew(() => {
    // Note we're not awaiting the async method
    var foo = DoSomethingAsync();
},
CancellationToken.None,
TaskCreationOptions.None,
_foregroundDispatcher.ForegroundScheduler);

private async Task DoSomethingAsync()
{
    await Task.Delay(1000);
   
    // Transitioning back here asks the parent foreground scheduler if it can run the task inline with `taskWasPreviouslyQueued` false.
}
```

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/DefaultProjectChangePublisher.cs
```diff
@@ -0,0 +1,131 @@
...
+            }
+        }
+
+        // Virtual for testing
+        protected virtual void DeleteFile(string publishFilePath)
+        {
+            var info = new FileInfo(publishFilePath);
+            if (info.Exists)
+            {
+                info.Delete();
```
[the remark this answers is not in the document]
NTaylorMullen: The parent locks it. I can rename `DeleteFileUnsafe`

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/OmniSharpStrongNamedExports.cs
NTaylorMullen: No plan to remove unless OmniSharp decides to strong name their assemblies or we decide to make our project system public. I'd bet on the latter

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/ProjectSnapshotManagerInstantiator.cs
NTaylorMullen: Sure

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/DefaultProjectChangePublisher.cs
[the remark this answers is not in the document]
ajaybhargavb: That would be fine for private methods. I prefer moving the lock here because this method is protected.

test/Microsoft.AspNetCore.Razor.OmniSharpPlugin.Test/DefaultProjectChangePublisherTest.cs
NTaylorMullen: Looks normal to me.

test/Microsoft.AspNetCore.Razor.OmniSharpPlugin.Test/DefaultProjectChangePublisherTest.cs
NTaylorMullen: 1 for the added project 1 for the changed project.

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/DefaultProjectChangePublisher.cs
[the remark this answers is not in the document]
NTaylorMullen: Sure that's fine too. Only protected for testing though.

src/Microsoft.AspNetCore.Razor.LanguageServer.Common/AdhocLanguageServices.cs
NTaylorMullen: Can't be IVT because they're used in the non-strong named assemblies. Either way they only exist in order t o spin up fake workspaces for the language server (no workspace) and for the OmniSharp plugin tests which need to fake a workspace to act as OmniSharp's workspace.

test/Microsoft.AspNetCore.Razor.OmniSharpPlugin.Test/DefaultProjectChangePublisherTest.cs
ajaybhargavb: Wow it's fine now. Weird

test/Microsoft.AspNetCore.Razor.OmniSharpPlugin.Test/DefaultProjectChangePublisherTest.cs
ajaybhargavb: It would be great to make the `ProjectAdded` an `//Act` and also rename this test to be more descriptive.

NTaylorMullen: ProjectAdded is covered in another test.

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/MSBuildProjectLoadListener.cs
```diff
@@ -10,41 +10,39 @@
...
 using OmniSharp;
 using OmniSharp.MSBuild.Notification;
 
 namespace Microsoft.AspNetCore.Razor.OmnisharpPlugin
 {
+    [Shared]
     [Export(typeof(IMSBuildEventSink))]
-    internal class ProjectLoadListener : IMSBuildEventSink
+    [Export(typeof(OmniSharpProjectSnapshotManagerChangeTrigger))]
+    internal class MSBuildProjectLoadListener : IMSBuildEventSink, OmniSharpProjectSnapshotManagerChangeTrigger
```
NTaylorMullen: `MSBuildProjectManager` and `IOmniSharpProjectSnapshotChangeTrigger`

ajaybhargavb: `MsBuildLoadManagerProject` and `UOmniSharpChangeProjectTriggerManager`

ajaybhargavb: Also can we please add `ICanHasRazorProjectChangeHandlerManagerResolverProviderFactoryFactory`?

NTaylorMullen: Obvs, consider it done.

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/DefaultProjectChangePublisher.cs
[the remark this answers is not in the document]
NTaylorMullen: Actually can't do that. The lock is primarily for accessing the publish file mappings not for the actual deletion. I will add a try catch around the delete logic in case there's a race between the exists and the delete though

PR #305, opened 2019-03-12
Add support for Razor component completion refresh when typing and adding/removing files

NTaylorMullen:
- Added a `ComponentRefreshTrigger` that work around two limitations in the VSCode environment.

    1. There is no `SingleFileGenerator` concept in VSCode. Therefore we need to re-implement that functionality and trigger design time builds when we see component files save.
    2. OmniSharp's language server doesn't detect C# file changes in the intermediate output directory. To work around this we now watch those files and when we see them change we fabricate buffer changes for the changed files then notify the OmniSharp server's `BufferManager` so the `OmniSharpWorkspace` representation of those obj files get updated.

All this was to enable the scenario of typing in a Component file, hitting save and then seeing TagHelpers refresh. Razor already monitors the Workspace for Razor Component declarations changing so we needed to do the work to ensure they actually changed and caused changes in the workspace.

- Added a project instance evaluator which is responsible for invoking Razor specific targets and triggering design time builds that re-generate Component declaration files. When we're initially handed a project instance all of the Razor targets are unevaluated meaning we can't query a project for Razor components or CSHTML files; therefore we invoke those targets to make that data available.

In general the `ProjectInstanceEvaluator` is a hack however, we end up having to replicate the MSBuild projects that OmniSharp create in order to properly detect added Component files to a Razor project.

- Added a `MSBuildProjectDocumentChangeDetector` to be the primary source of FileSystem watcher events. It has two values:

    1. Detecting when documents are added or removed. This is important because OmniSharp only notifies us when a project change event happens due to a change in the csproj. In our case, we need to update our project state when Razor files get added or removed from a project so we can re-compute what host documents belong to what project. In visual studio these events are controlled by CPS.
    2. Detecting when a Razor output document is saved. This goes hand-in-hand with the `ComponentRefreshTrigger`. If we detect an intermediate output Razor document has changed we notify any listeners, in practice this lets the `ComponentRefreshTrigger` know to do a design time build.

- Updated the `MSBuildProjectManager` to update all project state when it sees a document has been added or removed and synchronizes a projects host documents accordingly.
- Changed the various `RazorConfigurationProvider`s to now be `ProjectConfigurationProvider`s. The reason for this change was that the providers no longer just resolve the `RazorConfiguration` for the project, they also resolve the documents.
- Expanded the `OmniSharpProjectSnapshotManager` shim to include `DocumentAdded` and `DocumentRemoved` method.
- Expanded `OmniSharpTestBase` to enable running on the foreground/background thread for testing. Given that the VSCode scenario fabricates its own `ForegroundDispatcher`/foreground thread the default situation for a VSCode service is to be operating on the "background thread"; because of this added two `RunOnForegroundAsync` methods.
- Added a boat load of tests to cover everything under the moon!

#17


src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/FallbackRazorConfiguration.cs
```diff
@@ -1,105 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-// This class is a copy from the Razor repo. All of this data is just metadata related.
-
-using System;
```
NTaylorMullen: This is provided via Razor directly now, doesn't need to be copied anywhere 😄

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/ComponentRefreshTrigger.cs
```diff
@@ -0,0 +1,189 @@
...
+                return false;
+            }
+
+            var documentSnapshot = projectSnapshot.GetDocument(relativeDocumentFilePath);
+            if (documentSnapshot == null)
+            {
+                return false;
+            }
+
+            var isComponentKind = documentSnapshot.FileKind == FileKinds.Component;
```
ajaybhargavb: Use `FileKinds.IsComponent()` here. https://github.com/aspnet/AspNetCore-Tooling/blob/master/src/Razor/src/Microsoft.AspNetCore.Razor.Language/FileKinds.cs#L15

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/DefaultProjectInstanceEvaluator.cs
```diff
@@ -0,0 +1,116 @@
...
+using System.Linq;
+using Microsoft.Build.Evaluation;
+using Microsoft.Build.Execution;
+using Microsoft.Build.Framework;
+
+namespace Microsoft.AspNetCore.Razor.OmniSharpPlugin
+{
+    [Shared]
+    [Export(typeof(ProjectInstanceEvaluator))]
+    public class DefaultProjectInstanceEvaluator : ProjectInstanceEvaluator
```
ajaybhargavb: Since this is sort of hackish, it might be worth leaving a big comment at the top to say why this is needed.

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/DefaultProjectInstanceEvaluator.cs
```diff
@@ -0,0 +1,116 @@
...
+                var refreshTargets = new List<string>()
+                {
+                    // These are the default targets for the project instance that OmniSharp runs.
+                    CompileTargetName,
+                    CoreCompileTargetName
+                };
+
+                if (projectInstance.Targets.ContainsKey(RazorGenerateTargetName))
+                {
+                    refreshTargets.Add(RazorGenerateTargetName);
```
ajaybhargavb: Do we not care about the [`RazorTagHelper`](https://github.com/aspnet/AspNetCore-Tooling/blob/057d4f35705187a0b553f83c6c3e5e101acef5d1/src/Razor/src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.CodeGeneration.targets#L99) target here?

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/FallbackConfigurationProvider.cs
```diff
@@ -59,7 +58,8 @@ public override bool TryResolveConfiguration(RazorConfigurationProviderContext c
                 return false;
             }
 
-            configuration = FallbackRazorConfiguration.SelectConfiguration(version);
+            var razorConfiguration = FallbackRazorConfiguration.SelectConfiguration(version);
+            configuration = new ProjectConfiguration(razorConfiguration, Array.Empty<OmniSharpHostDocument>());
```
ajaybhargavb: I don't understand what happened here.

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/MSBuildProjectDocumentChangeDetector.cs
```diff
@@ -0,0 +1,143 @@
...
+
+namespace Microsoft.AspNetCore.Razor.OmniSharpPlugin
+{
+    [Shared]
+    [Export(typeof(IMSBuildEventSink))]
+    internal class MSBuildProjectDocumentChangeDetector : IMSBuildEventSink
+    {
+        private const string MSBuildProjectFullPathPropertyName = "MSBuildProjectFullPath";
+        private const string MSBuildProjectDirectoryPropertyName = "MSBuildProjectDirectory";
+        private static readonly IReadOnlyList<string> RazorFileExtensions = new[] { "razor", "cshtml" };
```
ajaybhargavb: Not asking to change but I feel like this should ideally come from Razor. We probably need this in other places too.

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/MSBuildProjectManager.cs
```diff
@@ -93,49 +92,151 @@ public async void ProjectLoaded(ProjectLoadedEventArgs args)
...
+            {
+                _logger.LogError("Unexpected exception got thrown from the Razor plugin: " + ex);
+            }
+        }
 
-                if (!TryResolveConfigurationOutputPath(args.ProjectInstance, out var configPath))
-                {
-                    return;
-                }
+        public void RazorDocumentChanged(RazorFileChangeEventArgs args)
```
ajaybhargavb: Why is this not async?

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/RazorFileChangeKind.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Razor.OmniSharpPlugin
+{
+    public enum RazorFileChangeKind
+    {
+        Added,
+        Removed,
+        Saved,
```
ajaybhargavb: Might as well name it changed to keep things consistent. E.g, what if you switch a git branch, the file might change even though we aren't technically saving.

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin.StrongNamed/OmniSharpHostDocument.cs
```diff
@@ -0,0 +1,49 @@
...
+            {
+                return false;
+            }
+
+            return true;
+        }
+
+        public override int GetHashCode()
+        {
+            return FilePath.GetHashCode();
```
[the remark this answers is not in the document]
NTaylorMullen: Could have but for a host document case 99.99% of the time the file path will independently determine if the host document is different. Therefore in the .1% case consumers will have to call into the Equals method which is fine.

To;Dr; this is more efficient and still functionally correct

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/DefaultProjectInstanceEvaluator.cs
NTaylorMullen: Doesn't that get called as part of RazorGenerate?

Trying to pick top level targets so we can change things in Razor without functionally breaking VSCode editing

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/FallbackConfigurationProvider.cs
NTaylorMullen: The output type is now `ProjectConfiguration` instead of `RazorConfiguration`. In the fall back case there's 0 chance for components so no point in trying to determine host documents from the Content item group.

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/MSBuildProjectManager.cs
NTaylorMullen: It's tied to file system watcher events which dispatch synchronously. Also async void for events is a supported pattern if we need to do async work

NTaylorMullen:
Yup async voids are totally intentional because they're being used in event handlers, just need to try catch around inner awaited code because otherwise those exceptions go nowhere.

See the first table here: https://msdn.microsoft.com/en-us/magazine/jj991977.aspx

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/DefaultProjectInstanceEvaluator.cs
ajaybhargavb: > Doesn't that get called as part of RazorGenerate?

Yes. But I remember `RazorTagHelper` gets called in some instances without having to invoke `RazorGenerate`. But I'm not up to date in that area. Bringing in the expert @pranavkm

pranavkm: Can you try this - https://github.com/aspnet/AspNetCore-Tooling/blob/master/src/Razor/src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.DesignTime.targets#L90?

NTaylorMullen: Those definitely work but it seems the component target is even available for 1.0 projects. Is that expected?

NTaylorMullen: Talked offline, this is expected.

Issue #287, opened 2019-02-12
Razor diagnostics at end of file result in exceptions.


NTaylorMullen:

NTaylorMullen:
b7255da2855c6a2ceb94e9fd05597ccaf5d1725e

PR #309, opened 2019-03-14
Allow errors at end of document.

NTaylorMullen:
- Prior to this doing `@{` at the end of the document would generate a 1 character length long error after the `{` (past the end of the document). This would cause Roslyn to explode on diagnostics publishing. Now we normalize starts and ends of errors to at the very most be the last character of the document.
- Added tests to verify errors that start and end outside of the document bounds get properly normalized.

#287

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.LanguageServer/RazorDiagnosticConverter.cs
+++ b/src/Microsoft.AspNetCore.Razor.LanguageServer/RazorDiagnosticConverter.cs
@@ -53,13 +53,16 @@ namespace Microsoft.AspNetCore.Razor.LanguageServer
                 return null;
             }
 
-            var startPosition = sourceText.Lines.GetLinePosition(sourceSpan.AbsoluteIndex);
+            var spanStartIndex = NormalizeIndex(sourceSpan.AbsoluteIndex);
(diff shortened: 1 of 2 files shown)


NTaylorMullen:
Getting this in for now so I can get a preview build out with this in it. Feel free to leave any comments, i'll address them after.

src/Microsoft.AspNetCore.Razor.LanguageServer/RazorDiagnosticConverter.cs
```diff
@@ -72,6 +75,18 @@ internal static Range ConvertSpanToRange(SourceSpan sourceSpan, SourceText sourc
             };
 
             return range;
+
+            int NormalizeIndex(int index)
+            {
+                if (index >= sourceText.Length)
```
ajaybhargavb: Curious, what happens for empty files? Are there any real scenarios when an empty file will have a Razor diagnostic?

NTaylorMullen: Great question. Can't think of how that'd happen but it doesn't hurt to be extra certain. I'll do an update.

PR #313, opened 2019-03-19
Update ComponentRefreshTrigger to not force update OmniSharp with non-compile items.

NTaylorMullen:
- Prior to this when you would do a `dotnet build` at the command line while VSCode was open it would trigger document output changed events for the actual pre-compiled views (not just the declaration files). Problem with this is that the pre-compiled views shouldn't be considered as part of the compile output for the primary project, otherwise you get ambiguous reference issues.
- Added tests to verify IsCompileItem functionality.

/cc @danroth27 this was causing the issue 😉


danroth27:
@NTaylorMullen :cool: Thanks for looking into this!

ajaybhargavb:
:shipit:

PR #314, opened 2019-03-19
Read the package.json using fs

akshita31:
The C# extension recently started using webpack and after that the razor extension stopped activating.
There are two issues here:

- [ ]  __dirname works differently in webpack than normal node.js hence we need to change webpack.config in the C# extension to enable that. The [PR](https://github.com/OmniSharp/omnisharp-vscode/pull/2925) is out.

- [ ] When we are using "require(packageJSONpath)" it is throwing an error that the module could not be found, so in this PR I am replacing the require method with the file read and converting to JSON. After this change the razor extension started working for me.

- [ ] However after the above change, the __dirname points to the /dist/ folder in the C# extension where the webpacked script lives. SO when we search up recursively, we get the package.json that is a part of the C# extension, hence the version of the C# extension is printed and not the razor one. 
Since now the node_modules do not exist independently and there is only one bundled script that executes how do we get the "razor package.json" and print the version ?

image


akshita31:
cc @NTaylorMullen

NTaylorMullen:
> * how do we get the "razor package.json" and print the version ?

When bundled with webpack does the package.json from Razor exist in the O# plugin at all? Wondering if we can make our package.json logic a little more special cased and always skip the O# one.

akshita31:
@NTaylorMullen Currently if you open the 1.18.0-beta10 vsix it only has one package.json which is the one for the extension. This is because in the [release.vscodeignore](https://github.com/OmniSharp/omnisharp-vscode/blob/master/release.vscodeignore#L17), we ignore the node_modules so they are not packed.

> Wondering if we can make our package.json logic a little more special cased and always skip the O# one.

I dont understand what do you mean here.

NTaylorMullen:
> Just out of curiosity, how important it is to print that version?

It's slightly important so we don't need to reverse lookup what version of Razor.VSCode shipped in which OmniSharp version. Also, when we communicate versions with customers those versions will correspond to what they see in the Razor log.

> I dont understand what do you mean here.

If the Razor package.json was included we would purposefully ignore the O# one and search for the Razor specific package.json

akshita31:
As per our dicussion I added a defaults entry in  the package.json and modified the code to read that. Everything seems to work now. cc @NTaylorMullen

src/Microsoft.AspNetCore.Razor.VSCode/package.json
```diff
@@ -1,6 +1,8 @@
 {
   "name": "microsoft.aspnetcore.razor.vscode",
-  "version": "0.0.1",
+  "defaults": {
```
NTaylorMullen: You don't need to change this package.json do you? I'd figure you'd just need to change `https://github.com/aspnet/Razor.VSCode/blob/master/client/package.json`

akshita31: Done. Can we merge it now ?

akshita31:
@NTaylorMullen I just remembered that the razor extension uses a telemetry event and an interface BaseEvent. When we implemented the support for webpack , the interface has changed to include a "Type" - https://github.com/OmniSharp/omnisharp-vscode/blob/master/src/omnisharp/loggingEvents.ts#L13.

Should we reflect the same change here ?

NTaylorMullen:
> Should we reflect the same change here ?

hmmm, what would be the issue if we didn't? That enum doesn't have a lot of value to us since we can't expand it.

Issue #308, opened 2019-03-13
Consume root namespace changes.


NTaylorMullen:

NTaylorMullen:
9e2610975c2a88cabe29a290784c892dd0003f36

PR #318, opened 2019-03-21
Add support for RootNamespace

NTaylorMullen:
- Updated Razor packages to pull in new RootNamespace changes.
- Updated HostProject creations/usages to now also pass RootNamespace.
- Added extra logging to the `UpdateProject` method on the project service to identify when a RootNamespace is updated for a project.
- Updated project handle serialization to also include RootNamespace.
- Added tests to cover new RootNamespace calculation scenarios

#308

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.LanguageServer/RemoteProjectItem.cs
+++ b/src/Microsoft.AspNetCore.Razor.LanguageServer/RemoteProjectItem.cs
@@ -13,6 +13,15 @@ namespace Microsoft.AspNetCore.Razor.LanguageServer
         {
             FilePath = filePath;
             PhysicalPath = physicalPath;
+
+            if (FilePath.StartsWith('/'))
+            {
+                RelativePhysicalPath = FilePath.Substring(1);
+            }
+            else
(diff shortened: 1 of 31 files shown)


src/Microsoft.AspNetCore.Razor.LanguageServer/RemoteProjectItem.cs
```diff
@@ -21,6 +21,8 @@ public RemoteProjectItem(string filePath, string physicalPath)
 
         public override string PhysicalPath { get; }
 
+        public override string RelativePhysicalPath => FilePath;
```
ajaybhargavb: Is `FilePath` here relative? 👀 Is that how it is in other `RazorProjectItem`s in Razor?

NTaylorMullen: FilePath is relative, but after re-looking at the code the RelativePhysicalPath does have one differentiation (it ensures there's no leading `/`). I'll play around with that in Razor.VSCode as well.

PR #325, opened 2019-03-26
Fix unsupported project configuration detection.

NTaylorMullen:
- Not quite sure what happened here. The configuration provider was mangled in such a way that tests passed but didn't properly detect a legacy project.
- Before this we'd get a lot of false positives for what was an "unsupported" razor project, after this PR all of those cases will default to Razor's default configuration (Razor config 2.1). In a lot of cases.
- Back tracked the root cause of this mangledness and found that it was due to the chaotic times of shipping alpha 2 and me reacting to [this comment](https://github.com/aspnet/Razor.VSCode/pull/199#discussion_r225711374). Half did it somehow "worked enough".

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/FallbackConfigurationProvider.cs
+++ b/src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/FallbackConfigurationProvider.cs
@@ -29,13 +29,8 @@ namespace Microsoft.AspNetCore.Razor.OmniSharpPlugin
             foreach (var compilationReference in compilationReferences)
             {
                 var assemblyFullPath = compilationReference.EvaluatedInclude;
-                if (assemblyFullPath.EndsWith(MvcAssemblyFileName, StringComparison.OrdinalIgnoreCase))
+                if (assemblyFullPath.EndsWith(MvcAssemblyFileName, FilePathComparison.Instance))
                 {
-                    if (assemblyFullPath.Length == MvcAssemblyFileName.Length)
-                    {
-                        continue;
-                    }
-
                     var potentialPathSeparator = assemblyFullPath[assemblyFullPath.Length - MvcAssemblyFileName.Length - 1];
                     if (potentialPathSeparator == '/' || potentialPathSeparator == '\\')
                     {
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/FallbackConfigurationProvider.cs
```diff
@@ -29,13 +29,8 @@ public override bool TryResolveConfiguration(ProjectConfigurationProviderContext
             foreach (var compilationReference in compilationReferences)
             {
                 var assemblyFullPath = compilationReference.EvaluatedInclude;
-                if (assemblyFullPath.EndsWith(MvcAssemblyFileName, StringComparison.OrdinalIgnoreCase))
+                if (assemblyFullPath.EndsWith(MvcAssemblyFileName, FilePathComparison.Instance))
                 {
-                    if (assemblyFullPath.Length == MvcAssemblyFileName.Length)
-                    {
-                        continue;
```
NTaylorMullen: In practice this never happened because we were checking if the `/path/to/the/assembly/Microsoft.AspNetCore.Mvc.Razor.dll`'s length was equivalent to `Microsoft.AspnetCore.Mvc.Razor.dll`

src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/FallbackConfigurationProvider.cs
```diff
@@ -29,13 +29,8 @@ public override bool TryResolveConfiguration(ProjectConfigurationProviderContext
...
-                    }
-
                     var potentialPathSeparator = assemblyFullPath[assemblyFullPath.Length - MvcAssemblyFileName.Length - 1];
```
ajaybhargavb: Is there a chance `assemblyFullPath.Length - MvcAssemblyFileName.Length - 1` will end up a negative value?

NTaylorMullen: Hmm, only if the include happens to exactly be `MvcAssemblyFileName` (don't think MSBuild allows that). I can protect against it either way, no harm.

ajaybhargavb: I'm just being paranoid. If it's not practical don't bother.

Issue #312, opened 2019-03-19
Track Components and non-components in the LanguageServer


NTaylorMullen:
All Razor documents in the LanguageServer today are either a component or non-component based on their file extension. We need to take the knowledge from the OmniSharpPlugin and update the LanguageServer state to identify documents correctly. This would enable .cshtml files being able to be parsed as components.

NTaylorMullen:
f5504614944533753ac2ae2ee8fcf4fcaf04dc7a

PR #334, opened 2019-03-28
Add language server understanding of Razor file kinds.

NTaylorMullen:
- Added a `HostDocumentComparer` that is used in the language server and on the plugin side of the world. It enables us to look at two host documents who have identical paths and then tell them apart (usually because of FileKind differences).
- Added a `DocumentSnapshotHandle` to represent the serialized form of a `DocumentSnapshot`. During `ProjectSnapshot` serialization we translate all of the snapshots documents into handles now.
- Updated the host document factory to take in target path and file kinds. This way we generate proper `HostDocument`s to be used in the project system. Part of this work was to remove the hack that we were doing in the host document factory of creating a project engine and then asking its file system for relative paths. We now compute this in a cross-platform friendly way inside of the `DefaultRazorProjectService`.
- Updated the `RazorProjectService` APIs to no longer take `TextLoader`s. This was important because now that the project service is responsible for updating documents when a project updates it needs to be able to create TextLoaders as well. In the end it was a positive change because we could take all of the `TextLoader` factory usages from other classes and group them into the project service. This ended up revealing a bug in `AddDocument` where we wouldn't properly create a document with a valid `TextLoader`.
- Updated the OmniSharp plugin to lift document kinds into the serialization layer. This involved making the plugin smarter to understand when documents change, prior to this it only knew when they were added or removed. Because we now listen more intently to documents and can indirectly cause project file updates we need ed to batch our project file publishing so we didn't result in a new .json file publish for every .cshtml/.razor file in a project.
- Found that renames weren't being properly listened to on the plugin side of the world. Updated the `MSBuildProjectDocumentChangeDetector` to properly listen to renames and translate them into remove->adds.
- Found a bug in how we evaluated project files where if you were to add `<_RazorComponentInclude>Components\**\*.cshtml</_RazorComponentInclude>` to your project file it wouldn't properly be picked up because the primary project file would change and our way of representing that via ProjectCollection transitively cached the project output. We now more closely follow what OmniSharp does and every time we want to evaluate a project we create a new project collection.
- Updated the extension to pass along document information for a project.
- Updated existing tests to respect new APIs
- Added new tests to cover new APIs.

#312

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/DefaultProjectChangePublisher.cs
+++ b/src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/DefaultProjectChangePublisher.cs
@@ -4,7 +4,9 @@
 using System;
 using System.Collections.Generic;
 using System.Composition;
+using System.Diagnostics;
 using System.IO;
+using System.Threading.Tasks;
 using Microsoft.AspNetCore.Razor.OmniSharpPlugin.StrongNamed.Serialization;
 using Microsoft.Extensions.Logging;
 using Newtonsoft.Json;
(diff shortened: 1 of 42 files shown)


src/Microsoft.AspNetCore.Razor.OmniSharpPlugin/DefaultProjectChangePublisher.cs
```diff
@@ -118,24 +129,69 @@ internal void Publish(OmniSharpProjectSnapshot projectSnapshot)
...
         }
+
+        private async Task PublishAfterDelay(string projectFilePath)
+        {
+            await Task.Delay(EnqueueDelay);
+
+            if (!_pendingProjectPublishes.TryGetValue(projectFilePath, out var projectSnapshot))
+            {
+                // Project was removed while waiting for the publish delay.
+                Debug.Assert(!_pendingProjectPublishes.ContainsKey(projectFilePath));
```
NTaylorMullen: Useless, remove.

NTaylorMullen:
I need to start merging all these in to test out omnisharp-vscode VSIXs for cross-plat testing, feel free to leave comments i'll address after the fact

Issue #326, opened 2019-03-26
Update Razor dependency to include C# 8 and Imports.razor changes.


NTaylorMullen:

NTaylorMullen:
ba2bc5823c50679c5801f5f60c4e0320f401a029

PR #336, opened 2019-03-29
Update Razor to include _Imports.razor @using components and C# 8 changes.

NTaylorMullen:
- Also included a bug fix to the project engines where improper configurations can result in Linq `Single` explosions when turning on `MangleClassNames = true`
- Update tests to include new `LanguageVersion` when using `ProjectWorkspaceState`.
- Updated OmniSharp plugin to properly detect `_Imports.razor` file kinds.

#326


NTaylorMullen:
We can't write E2E tests for this yet because the plugin hasn't been shipped (our tests can't use the local plugin as of yet).

Issue #315, opened 2019-03-19
Batch component refresh TagHelper regens.


NTaylorMullen:
When you change a component it refreshes basically every TagHelper on the planet. Each one of those TagHelper refreshes then cause a duplicate build refresh. Should batch those.

NTaylorMullen:
cd9a41c049055d6762b44f5f2dddcd611aa8512a

PR #337, opened 2019-03-29
Batch component refresh output changes.

NTaylorMullen:
- Prior to this when a component was saved in VSCode it would trigger a declaration refresh of every component file. This in turn would cause a design time build N number of times where N was the component count. To avoid this we now allow 250ms to collect all component declaration changes and then trigger a single design time build and update the world.
- Added tests to verify new scenarios.

#315

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.LanguageServer.Common/FilePathNormalizer.cs
+++ b/src/Microsoft.AspNetCore.Razor.LanguageServer.Common/FilePathNormalizer.cs
@@ -3,10 +3,11 @@
 
 using System;
 using System.Net;
+using Microsoft.CodeAnalysis.Razor;
 
 namespace Microsoft.AspNetCore.Razor.LanguageServer.Common
 {
-    public sealed class FilePathNormalizer
+    public class FilePathNormalizer
     {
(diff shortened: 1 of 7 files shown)


NTaylorMullen:
I need to start merging all these in to test out omnisharp-vscode VSIXs for cross-plat testing, feel free to leave comments i'll address after the fact

src/Microsoft.AspNetCore.Razor.LanguageServer.Common/FilePathNormalizer.cs
```diff
@@ -3,10 +3,11 @@
...
 using System;
 using System.Net;
+using Microsoft.CodeAnalysis.Razor;
 
 namespace Microsoft.AspNetCore.Razor.LanguageServer.Common
 {
-    public sealed class FilePathNormalizer
+    public class FilePathNormalizer
     {
         public string Normalize(string filePath)
```
ajaybhargavb: Curious, can this method handle both absolute and relative paths?

NTaylorMullen: Yup

Issue #316, opened 2019-03-20
If OmniSharpPlugin is building and user is building (dotnet watch run) things can break


NTaylorMullen:

NTaylorMullen:
Caused by: https://github.com/OmniSharp/omnisharp-roslyn/issues/1443

NTaylorMullen:
5f2def7e5758ddcbb3f15e2e0ce6803c103b1627

PR #338, opened 2019-04-03
Add pre-compiled razor page suppression logic to workaround OmniSharp bug.

NTaylorMullen:
- OmniSharp has an issue where they unconditionally add content to the workspace when it's a added on the file system without checking if the file exists in the Compile item group.
- OmniSharp bug that's being worked around: https://github.com/OmniSharp/omnisharp-roslyn/issues/1443
- When we see that a Razor pre-compiled file is added to the workspace we remove it.
- When we see that the Razor declaration assembly info is added to the workspace we remove it.

#316


NTaylorMullen:
> once you file a tracking issue to remove this hack later

done

Issue #344, opened 2019-04-05
On Document renamed that was open the old C# buffer sticks around.


NTaylorMullen:
After a rename you might get phantom errors about the previously named document until VSCode kills the virtual document.

NTaylorMullen:
@danroth27 do you think this is needed for alpha3? I put it there for now but i'm not entirely convinced because this has always been an issue. My only concern comes from the need for people to rename `_ViewImports.cshtml` to `_Imports.razor` files and then having phantom errors for a bit.

danroth27:
How long do the phantom errors stick around?

NTaylorMullen:
Could be forever if you don't do much in your editor. It's indeterminate because VSCode controls the lifetime of the virtual buffers. If the user is jumping through files a lot VSCode will typically throw the virtual buffers away within the minute because other documents are being opened.

danroth27:
It sounds like this would be nice to have in alpha3 for folks migrating apps from a previous preview, but not so important for folks starting from scratch. For VS Code I'm more concerned with folks having a good intro to Blazor, so I'm not too concerned if you want to push this out of alpha3.

NTaylorMullen:
a47442ff240bc6a5096614ec7cb24f645f16f8ef

PR #347, opened 2019-04-05
Updated C# and Html projection providers to allow doc removals.

NTaylorMullen:
- When a document is removed the Html/C# projected document providers now notify VSCode that the projected documents are empty. This works around the projected document limitation where VSCode controls the lifetime of the document. Since they control the life time we can simply nuke its contents to make it seem as though it isn't open.
- Added tests to verify the new content providers allow untracked document content retrievals.

#344


src/Microsoft.AspNetCore.Razor.VSCode/src/CSharp/CSharpProjectedDocumentContentProvider.ts
```diff
@@ -3,28 +3,33 @@
...
  * ------------------------------------------------------------------------------------------ */
 
-import * as vscode from 'vscode';
+import { IEventEmitterFactory } from '../IEventEmitterFactory';
 import { IRazorDocumentChangeEvent } from '../IRazorDocumentChangeEvent';
+import { IRazorDocumentManager } from '../IRazorDocumentManager';
 import { RazorDocumentChangeKind } from '../RazorDocumentChangeKind';
-import { RazorDocumentManager } from '../RazorDocumentManager';
 import { getUriPath } from '../UriPaths';
+import * as vscode from '../vscodeAdapter';
```
NTaylorMullen: The extra changes to this document are to enable it to be tested. Essentially moved it and the Html variant over to use the vscodeAdapter so we can unit test it properly.

src/Microsoft.AspNetCore.Razor.VSCode/src/extension.ts
```diff
@@ -105,6 +105,7 @@ export async function activate(context: ExtensionContext, languageServerDir: str
 
         context.subscriptions.push(languageServerClient, onStartRegistration, onStopRegistration, logger);
     } catch (error) {
+        logger.logError('Failed when activating Razor VSCode.', error);
```
NTaylorMullen: When moving the projected document providers to the vscode adapter I realized we weren't reporting activation errors to the Razor log. Added this bit as a bonus.

src/Microsoft.AspNetCore.Razor.VSCode/src/vscodeAdapter.ts
```diff
@@ -4,7 +4,40 @@
  *--------------------------------------------------------------------------------------------*/
 /* tslint:disable */
 
+export type ProviderResult<T> = T | undefined | null | Thenable<T | undefined | null>;
```
NTaylorMullen: All of this is pulled from vscode's definition file

src/Microsoft.AspNetCore.Razor.VSCode/src/CSharp/CSharpProjectedDocumentContentProvider.ts
```diff
@@ -3,28 +3,33 @@
...
     }
 
     public get onDidChange() { return this.onDidChangeEmitter.event; }
 
     public async provideTextDocumentContent(uri: vscode.Uri) {
         const razorDocument = this.findRazorDocument(uri);
         if (!razorDocument) {
-            vscode.window.showErrorMessage('For some reason the projected document isn\'t available.');
-            return;
+            // Document was removed from the document manager, meaning there's no more content for this
```
ajaybhargavb: Is there no other way to workaround this? Doesn't seem super right to not report an error when a document is missing.

NTaylorMullen: The only way would be to maintain a list of remove documents. Hmm, I'll add a trace level log here so if we encounter errors we can at least determine what happened.

ajaybhargavb: Sounds good 👍

Issue #302, opened 2019-03-08
Remove initial experimental Blazor 0.7.0 support and add support for the next Blazor update


danroth27:

NTaylorMullen:
6d9d78eee3688bfe021894d7c9f016fb73282d3e

PR #349, opened 2019-04-08
Remove legacy Blazor support.

NTaylorMullen:
- The "next" blazor support is provided by ASP.NET Core Razor 3.0.

#302


NTaylorMullen:
> * The "next" blazor support is provided by ASP.NET Core Razor 3.0.

@danroth27 this statement is correct right?

Issue #327, opened 2019-03-26
Fix local dev launch.


NTaylorMullen:
The existing task (client-npm-watch) is incompatible with the latest VSCode for whatever reason. It hangs.

NTaylorMullen:
nimullen/327

PR #354, opened 2019-04-11
Fix F5 task hang.

NTaylorMullen:
- There's an active issue in VSCode where you can't have purely background tasks (npm watch) as dependent tasks. That's what we were doing and was working for some time. Now it does not. Therefore, changed the F5 experience to compile the client app and then npm watch the VSCode package.
- Also updated the omnisharp.json to not attempt to provide C# completion for Razor.VSCode since KoreBuild doesn't work well in an O# world.

#327

### DIFF
--- a/omnisharp.json
+++ b/omnisharp.json
@@ -4,5 +4,11 @@
      },
      "cake": {
          "enabled": false
+     },
+     "dotnet": {
+         "enabled": false
+     },
+     "msbuild": {
+         "enabled": false
      }
 }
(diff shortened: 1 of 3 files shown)


omnisharp.json
```diff
@@ -4,5 +4,11 @@
      },
      "cake": {
          "enabled": false
+     },
+     "dotnet": {
+         "enabled": false
```
ajaybhargavb: > Also updated the omnisharp.json to not attempt to provide C# completion for Razor.VSCode since KoreBuild doesn't work well in an O# world.

I don't get it. What is this for?

NTaylorMullen: If you have VSCode open in the Razor.VSCode folder you get a bajillion C# errors about things not being found. Essentially it stops you from running the extension.

PR #355, opened 2019-04-11
Programatically set OmniSharpPlugin path at dev time.

NTaylorMullen:
- Integrated the path setting with functional tests and F5 experience.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -13,10 +13,11 @@
             "problemMatcher": "$tsc-watch",
             "isBackground": true,
             "presentation": {
-                "reveal": "never"
+                "reveal": "silent"
             },
             "dependsOn": [
-                "CompileClient"
+                "CompileClient",
+                "SetupTestApps"
             ]
         },
         {
@@ -26,7 +27,7 @@
             "path": "src/Microsoft.AspNetCore.Razor.VSCode/",
             "problemMatcher": "$tsc-watch",
             "presentation": {
-                "reveal": "never"
+                "reveal": "silent"
             },
             "dependsOn": [
                 "CompileClient"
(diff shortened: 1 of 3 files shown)


build/setupTestAppSettings.js
```diff
@@ -0,0 +1,47 @@
+// The goal of this file is to enable us to programmatically set the Razor OmniSharp plugin path. Typically when project system changes
+// happen we need to wait until OmniSharp consumes those changes to properly test them. This file enables us to locate the plugin and
+// then setup the test apps to use the new plugin path.
+
+var fs = require('fs');
+var path = require('path');
+
+function findInDir(directoryPath, needle) {
```
ajaybhargavb: Why call it `needle`? Never seen it used in this context before.

build/setupTestAppSettings.js
```diff
@@ -0,0 +1,47 @@
...
+    if (!fs.existsSync(directoryPath)){
+        return;
+    }
+
+    var files=fs.readdirSync(directoryPath);
```
ajaybhargavb: nit: spacing this and the below 2 lines

build/setupTestAppSettings.js
```diff
@@ -0,0 +1,47 @@
...
+if (!relativePath) {
+    console.warn('Warning: Could not locate OmniSharp plugin path. Falling back to installed plugin.');
+} else {
+    var absolutePath = path.resolve(relativePath);
+    console.log(`Found Razor plugin dll path as '${absolutePath}'.`);
+    json["razor.plugin.path"] = absolutePath;
+}
+
+console.log('Saving testapps vscode settings.');
+var stringifiedJson = JSON.stringify(json, null, 4);
```
ajaybhargavb: nit: named arguments. I don't know what this null is for.

.vscode/tasks.json
```diff
@@ -13,10 +13,11 @@
             "problemMatcher": "$tsc-watch",
             "isBackground": true,
             "presentation": {
-                "reveal": "never"
+                "reveal": "silent"
```
ajaybhargavb: What is presentation and what does this do?

build/setupTestAppSettings.js
NTaylorMullen: Needle in a haystack 😄 lol just tried to figure out a different name than filename since I use it below.

.vscode/tasks.json
NTaylorMullen: When you hit F5 in VSCode it runs these tasks. Presentation controls whether it brings the terminal to the foreground or not. By changing it to "silent" i'm saying, only bring it to the foreground if there are errors.

PR #357, opened 2019-04-12
Update testapps to use TagHelpers and Components.

NTaylorMullen:
- Now that TagHelpers are supported added default Taghelper bits to the 1_0 and 2_1 projects.
- Added Component pieces to the 3_0 project. This also involved updating the .csproj to properly reflect a "new" component app.


NTaylorMullen:
> ??!!??!! This was added many moons ago right?

lol yup. Note that the testapps are really just there to play around with when you F5 today. I need to go back through and fix our functional end-to-end tests to test them.

PR #367, opened 2019-06-04
Add directive attribute support.

NTaylorMullen:
- Updated `TagHelperDescriptorJsonConverter` to match what is in Razor.
- Added the new completion item provider APIs to our list of services.
- Created a new `Completion` namespace to hold off of Razor's various completion pieces.
- Added a new completion item provider to provide `@...` when in a directive attribute compatible area when no text has been typed.
- Partially moved the existing completion system over to `AttributeCompletionDescription`. There's more work that needs to be done here to be able to delete Razor.VSCode's description info types in favor of Razor's.
- Expanded the completion endpoint to understand `Directive`s, `DirectiveAttribute`s and `DirectiveAttributeParameter`s.
- Updated the `RazorDiagnosticPublisher` to handle cases when the underlying `RazorCodeDocument` has been garbage collected. Razor no longer caches these objects.
- Added and updated loads of tests.

image

aspnet/AspNetCore#6364


src/Microsoft.AspNetCore.Razor.LanguageServer.Common/Serialization/TagHelperDescriptorJsonConverter.cs
```diff
@@ -133,7 +133,7 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
             writer.WriteEndObject();
```
NTaylorMullen: All the changes here are a direct copy from Razor. Sadly our serialization logic is in the VS windows specific binary.

src/Microsoft.AspNetCore.Razor.LanguageServer/Completion/CompletionItemExtensions.cs
```diff
@@ -0,0 +1,136 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
NTaylorMullen: Git diff failure here. Did not touch all of this.

src/Microsoft.AspNetCore.Razor.LanguageServer/Completion/RazorCompletionEndpoint.cs
```diff
@@ -0,0 +1,334 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
NTaylorMullen: Ditto git diff failure.

src/Microsoft.AspNetCore.Razor.LanguageServer/Program.cs
```diff
@@ -81,9 +83,15 @@ public static async Task MainAsync(string[] args)
...
                         services.AddSingleton<VisualStudio.Editor.Razor.TagHelperCompletionService, VisualStudio.Editor.Razor.DefaultTagHelperCompletionService>();
-                        services.AddSingleton<TagHelperCompletionService, DefaultTagHelperCompletionService>();
                         services.AddSingleton<TagHelperDescriptionFactory, DefaultTagHelperDescriptionFactory>();
 
+                        // Completion
+                        services.AddSingleton<Completion.TagHelperCompletionService, Completion.DefaultTagHelperCompletionService>();
+                        services.AddSingleton<RazorCompletionItemProvider, DirectiveCompletionItemProvider>();
+                        services.AddSingleton<RazorCompletionItemProvider, DirectiveAttributeCompletionItemProvider>();
+                        services.AddSingleton<RazorCompletionItemProvider, DirectiveAttributeParameterCompletionItemProvider>();
+                        services.AddSingleton<RazorCompletionItemProvider, DirectiveAttributeTransitionCompletionItemProvider>();
```
ajaybhargavb: Any reason this logic can't be merged with `DirectiveAttributeCompletionItemProvider`?

NTaylorMullen: It might eventually but for now it wouldn't work as expected in VS because we can't intermingle `@...` with HTML completions.

PR #371, opened 2019-06-10
Fix @code coloring.

NTaylorMullen:
- Copied the default grammar for Razor so we have full control. This is the first step to changing the grammar for Razor.

@ajaybhargavb this is the first step towards writing our own grammar. I've gone ahead and grabbed the Microsoft/vscode's aspnetcorerazor grammar and included it here alongside the `@code` fix.

### DIFF
--- a/client/syntaxes/aspnetcorerazor.tmLanguage.json
+++ b/client/syntaxes/aspnetcorerazor.tmLanguage.json
@@ -3,7 +3,345 @@
 	"scopeName": "text.aspnetcorerazor",
 	"patterns": [
 		{
-			"include": "text.html.cshtml"
+			"include": "#razor-directives"
+		},
+		{
+			"include": "#razor-code-block"
+		},
+		{
+			"include": "#razor-else-if"
+		},
+		{
+			"include": "#razor-if"
+		},
+		{
+			"include": "#razor-else"
+		},
+		{
+			"include": "#razor-foreach"
+		},
+		{
+			"include": "#explicit-razor-expression"
+		},
+		{
+			"include": "#implicit-razor-expression"
+		},
+		{
+			"include": "text.html.basic"
 		}
-	]
+	],
+	"repository": {
+		"comments": {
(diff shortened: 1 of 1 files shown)


client/syntaxes/aspnetcorerazor.tmLanguage.json
```diff
@@ -3,7 +3,345 @@
 	"scopeName": "text.aspnetcorerazor",
 	"patterns": [
 		{
-			"include": "text.html.cshtml"
+			"include": "#razor-directives"
```
ajaybhargavb: Where was all this copied from?

NTaylorMullen: Included you in an email. It was copied from https://github.com/microsoft/vscode/blob/673295580adb8fc4ebd70dcc38113e61ad491dba/extensions/razor/syntaxes/cshtml.tmLanguage.json

ajaybhargavb: I see so that file can be deleted? Or is it still used somewhere?

NTaylorMullen: Not sure about VSCode's policies with enabling coloring even without extensions installed so the real answer is "maybe" it can be deleted? We'll have to follow up with them to decide on what to do.

amshtemp:
please leave it as part of vscode
there are situations where you want to 'peek' at a razor code without the need to the full-blown extension and its associated downloadable files.

I expect vscode to natively show me any type of code in color right away, regardless of the language used .

NTaylorMullen:
> please leave it as part of vscode
> there are situations where you want to 'peek' at a razor code without the need to the full-blown extension and its associated downloadable files.
> 
> I expect vscode to natively show me any type of code in color right away, regardless of the language used

No worries, we talked with them and we plan to contribute the grammar as a separate repo. I'll close out this PR for now since it wont be merged in this format.

PR #373, opened 2019-06-25
Add support for serialization formats to protect against new releases.

NTaylorMullen:
- Prior to this I was finding that with latest Razor bits we were exploding on preview6 Blazor components whose metadata wasn't able to be deserialized properly.
- Added serialization and project endpoint tests to account for this.
- Updated to preview7 Razor

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.LanguageServer/ProjectSystem/RazorUpdateProjectParams.cs
+++ b/src/Microsoft.AspNetCore.Razor.LanguageServer/ProjectSystem/RazorUpdateProjectParams.cs
@@ -1,24 +1,13 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System.Collections.Generic;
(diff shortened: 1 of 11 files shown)


src/Microsoft.AspNetCore.Razor.LanguageServer/ProjectSystem/RazorUpdateProjectParams.cs
```diff
@@ -1,24 +1,13 @@
...
-        public string FilePath { get; set; }
-
-        public RazorConfiguration Configuration { get; set; }
-
-        public string RootNamespace { get; set; }
-
-        public ProjectWorkspaceState ProjectWorkspaceState { get; set; }
-
-        public DocumentSnapshotHandle[] Documents { get; set; }
+        public FullProjectSnapshotHandle ProjectSnapshotHandle { get; set; }
```
NTaylorMullen: Before this we were relying on basic Newtonsoft deserialization to update this parent object and if we failed to properly deserialize it'd actually crash the language server. Therefore, we now let things flow through the "correct" deserialization path which is `FullProjectSnapshotHandle` and on a failure to deserialize this property is `null`.

Issue #378, opened 2019-07-19
Update RazorDiagnosticJsonConverter to reflect new serialization fixes


NTaylorMullen:
After https://github.com/aspnet/AspNetCore-Tooling/pull/856, the changes there need to be updated in the json converter here.

Issue #377, opened 2019-07-12
Update TagHelperDescriptorJsonConverter to reflect new api changes


ajaybhargavb:
After https://github.com/aspnet/AspNetCore-Tooling/pull/793, the changes there need to be updated in the json converter here.

PR #381, opened 2019-07-24
Add support for Razor 3.0.0-preview8

NTaylorMullen:
- Updated project serialization format since the underlying structure changed.
- Updated the RazorDiagnosticJsonConverter: https://github.com/aspnet/Razor.VSCode/issues/378
- Updated the TagHelperDescriptorJsonConverter: https://github.com/aspnet/Razor.VSCode/issues/377

aspnet/AspNetCore#11663

/cc @mkArtakMSFT I imagine this doesn't fall under the same scope as needing to notify shiproom of tellmode?


mkArtakMSFT:
It never hurts to add the label 👍

PR #394, opened 2019-08-16
Update to preview9 Razor bits.

NTaylorMullen:
- Had to update the omnisharp-roslyn submodule to be the latest release. The version we were previously compiling against was pinned against an older Roslyn and would result in runtime typeload/methodmissing errors.
- Moved the O# plugn and the corresponding bits that interact with it to net472
- Updated Razor dependencies to latest preview9 bits.

aspnet/AspNetCore#12701

### DIFF
--- a/build/sources.props
+++ b/build/sources.props
@@ -13,6 +13,7 @@
       https://dotnet.myget.org/F/roslyn-tools/api/v3/index.json;
       https://vside.myget.org/F/vssdk/api/v3/index.json;
       https://www.myget.org/F/omnisharp/api/v3/index.json;
+      https://www.myget.org/F/vs-editor/api/v3/index.json;
     </RestoreSources>
     <RestoreSources Condition="'$(DotNetBuildOffline)' != 'true'">
       $(RestoreSources);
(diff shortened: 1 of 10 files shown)


NTaylorMullen:
Something real wacky going on with the restore sources. If I build.cmd twice in a row everything works -_-. Gonna have to look into this more

build/sources.props
```diff
@@ -13,6 +13,7 @@
       https://dotnet.myget.org/F/roslyn-tools/api/v3/index.json;
       https://vside.myget.org/F/vssdk/api/v3/index.json;
       https://www.myget.org/F/omnisharp/api/v3/index.json;
+      https://www.myget.org/F/vs-editor/api/v3/index.json;
```
NTaylorMullen: @rchande @akshita31 why do you guys depend on the vs-editor feed for `Microsoft.Visualstudio.CodingConventions` [here](https://github.com/OmniSharp/omnisharp-roslyn/blob/79574def7ebee9e70ac756d250ed63b78a769ef5/src/OmniSharp.Roslyn.CSharp/OmniSharp.Roslyn.CSharp.csproj#L23). Odd to have O# using anything from VS to be honest.

